PR #19, opened 2016-11-14
adding casebook mapping value as ID for country list

smulvenna:


simonmoore100:
I'm getting a 'Cannot read property 'in_EU' of undefined' error originating from SendReturnOptionsController.js

PR #23, opened 2016-11-16
using casebook mapping value for application countries

smulvenna:


simonmoore100:
Looks fine and works as expected

PR #29, opened 2016-12-07
Added script to update database country list. First truncates then ad…

smulvenna:
…ds countries from scratch


simonmoore100:
Tested and verified. Merge when ready

PR #31, opened 2016-12-09
Updating SQL to change ’N/A’ to empty string for casebook submission

smulvenna:


simonmoore100:
Tested and verified. Merge when ready

PR #62, opened 2017-06-20
Add logging to try and identify problem when exporting app data

simonmoore100:


smulvenna:
Only additional log lines added. Looks good to me

smulvenna (approved):
Only log lines added. Looks good

PR #75, opened 2018-02-21
Use regex to only allow QR code generation for unique references

simonmoore100:


cmcgandy:
reviewed, tested and happy to merge

PR #77, opened 2018-04-13
Fix race condition whereby the app would think it has been submitted …

simonmoore100:
…before it actually would have.


cmcgandy:
reviewed and happy to merge

PR #78, opened 2018-04-19
if statement to check if the user has paid

cmcgandy:


simonmoore100 (approved):
Tested and works fine!

PR #84, opened 2018-08-22
Force appSubmittedStatus value to be false when starting a new app

simonmoore100:


cmcgandy:
reviewed and happy to merge

PR #86, opened 2018-08-29
SQL script to update document name as per KIM FCO0585

cmcgandy:
SQL script to change document name as per KIM FCO0585


simonmoore100 (changes requested):
The query itself works, but I would change the format of the document to be more like what you are replacing. e.g.

html_id shouldn't have any capital letters.
if html_id was 'criminal-records-bureau-crb-document' then this should become 'disclosure-and-barring-service-dbs-document'
The rest of the updates need to have the word 'document' at the end

simonmoore100 (approved):

PR #133, opened 2020-04-07
Feature/fco 138 cookie banner

martincarson:


config/routes.js
```diff
@@ -18,9 +18,12 @@ module.exports.routes = {
     '/terms-and-conditions'                         : { view: 'legalPages/terms' },
     '/privacy-policy'                               : { view: 'legalPages/privacy' },
     '/cookies'                                      : { view: 'legalPages/cookies' },
+    '/cookie-details'                              : { view: 'legalPages/cookiesDetails' },
+    '/cookies-opt-out'                              : { view: 'legalPages/cookiesOptOut' },
```
smulvenna, line 22: I don't think this view exists

smulvenna (changes requested):
Seems to work OK. I've added a note about a route that does't seem to exist

PR #161, opened 2020-06-10
DAC_Grouping_03: Remove unnecessary fieldset tag

simonmoore100:


martincarson (approved):
Aria-required has been removed too, but I think that is ok here. Happy to merge

PR #166, opened 2020-06-10
DAC_Unlabelled_forms_01 - named label correctly

martincarson:


cmcgandy (changes requested):
looks good Marty, although would you need to maybe apply the same place on the '/change-details' page. The input type="tel" id="phone" there.

martincarson:
@cmcgandy I think the change-details screen is for the user-service which is separate

cmcgandy (approved):
in that case, happy to approve

PR #167, opened 2020-06-12
removing the 'tabindex' to ensure there is no space when tabbing betw…

cmcgandy:
…een the last top search and select action


martincarson (approved):

martincarson (approved):
Tested, tabindex is removed and no issues occur on this screen

PR #175, opened 2020-06-24
DAC_Non-descriptive_Headings_01

cmcgandy:
Changing the title of page to from Select your documents to Search and select your documents


martincarson (approved):
Both text updates look good to me

PR #176, opened 2020-06-24
DAC_Non-descriptive_Labels_01

cmcgandy:
“skip this check” is changed to "skip the document check"

*need to edit paragraph sizing to ensure the '.' remains on the same line*


martincarson (approved):
Text change only. Looks good to me

PR #182, opened 2020-08-26
Re-branding changes

cmcgandy:
Current rebranding changes for Legalisation service


views/legalPages/terms.ejs
```diff
@@ -12,7 +12,7 @@
 
     <h2 class="font-medium">General</h2>
 
-    <p>These terms and conditions affect your rights and liabilities under the law. They govern your use of, and relationship with, the legalisation online service. They don’t apply to other services provided by the Foreign & Commonwealth Office (FCO), or to any other department or service which is linked to in this service.</p>
```
smulvenna, line 15: You have updated the abbreviation - Just need to update the full text

smulvenna (changes requested):

views/legalPages/terms.ejs
cmcgandy: good spot Sean, changes made

smulvenna (approved):
happy with the changes made

PR #186, opened 2020-08-27
Add 'opens in new tab' logo to link

simonmoore100:


sdmadden (approved):
As far as the change noted in the document goes, it all looks good. Now that this opens in the same tab, though, should the close button instead just be removed altogether?

You look to have changed its functionality to navigate backwards, but, for example, going:

- home page
- Cookies (in footer) (opens in new tab)
- Find out more about cookies
- close

Causes the whole tab to still close.

sdmadden (approved):

PR #244, opened 2021-06-07
FCDO-130: Limit premium service to 20 docs per application

simonmoore100:


assets/js/application_forms.js
```diff
@@ -183,12 +183,27 @@ $(document).ready(function(){
...
+    var isPremium = false
     $('body').find('.number').each(function () {
         sum += Number($(this).val());
         totalCost = sum * documentCost;
+
+        // Check if the 'max' attribute is defined for the doc quantity input
+        // if so it's a premium application
+        var maxAttr = $(this).attr('max');
+        if (typeof maxAttr !== 'undefined' && maxAttr !== false) {
+          isPremium = true;
```
benjamindi96k, line 195: This may always be true since max is defined on both non/premium document.ejs pages

views/businessForms/documentQuantity.ejs
```diff
@@ -58,24 +58,30 @@
...
+                    </span>
+
                     <input type="number" aria-required="true" step="any"
                            <% if (submit_status===true) { %>disabled="disabled"<% } %>
                            id="documentCount" name="documentCount"
                            value="<% if (form_values && !selected_docs_count) { %><%= (form_values.doc_count || '') %><% } else if (!form_values && selected_docs_count) { %><%= (selected_docs_count || '')%><% } %>"
-                           class="number govuk-input govuk-!-width-one-quarter"/>
+                           class="number govuk-input govuk-!-width-one-quarter"
+                            <% if (req.session.appType === 2) { %>
+                                max="20"
```
benjamindi96k, line 78: Perhaps just rename this (max="20") to maxPremDocs then change     
`var maxAttr = $(this).attr('max');    ` 
to
`var maxAttr = $(this).attr('maxPremDocs');`

benjamindi96k (commented):
Other than 2 comments all else looks good

benjamindi96k (approved):
Looks good to me now

assets/js/application_forms.js
simonmoore100: I've removed the max attribute from applicationForms/documentQuantity.ejs, since its only used when the flag to limit the standard applications is turned on. This is old functionality which we can remove in a cleanup exercise.

PR #255, opened 2021-07-20
Feature/eia 128

RichardBray:
# Description

 **[1 - FEATURE VIDEO RUN-THROUGH 🎨](https://drive.google.com/file/d/1nAmEy96mkXy5kkqchlJTDMSx4JTQS2hi/view?usp=sharing)**

 **[2 - CODE VIDEO RUN-THROUGH 💃](https://drive.google.com/file/d/138cKzZnEksasP6r2MVxD7iAZPxZBPqv_/view?usp=sharing)**

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-128

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:


api/controllers/ApplicationController.js
```diff
@@ -234,8 +234,12 @@ var applicationController = {
               if (application.serviceType == 1) {
                 console.log(id + " - displaying standard confirmation page to user");
                 return applicationController.confirmation(req, res);
-              }
-              else {
+              } else if (application.serviceType === 4) {
+                console.log(
+                    id + ' - displaying eApostille confirmation page to user'
+                );
```
jenbutongit: did you mean to keep this in? if so, you should use the sails logger (unless you've been told to do otherwise) since it will pipe the error messages in the right format to the right place

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -14,20 +17,30 @@ const CheckUploadedDocumentsController = {
...
+                )
+                    .then(() => {
+                        sails.log.info(
+                            `Url for document ${uploadedFile.filename} added to db`
+                        );
+                    })
+                    .catch((err) => {
+                        sails.log.error(err);
+                        res.serverError();
+                    });
```
superafroman: Appreciate you can't use async/await in a forEach callback, but this code is still asynchronous so will run in parallel to the rest of the function. If there is an error it's likely that `CheckUploadedDocumentsController._checkDocumentCountAndPaymentDetails` will already have been executed and `res.serverError();` would then fail (I'm assuming a few things here.. :) ) 

You've got a couple of options - chain the whole thing in promises or use a classic for loop with await (though this would require `renderPage` to be async, not sure if that's supported.

api/controllers/ApplicationTypeController.js
```diff
@@ -296,16 +294,29 @@ module.exports = {
...
+                                                    ) {
+                                                        return res.redirect(
+                                                            '/business-document-quantity?pk_campaign=DropOff-Service&pk_kwd=DropOff'
+                                                        );
+                                                    } else if (
+                                                        req.session.appType === 4
+                                                    ) {
+                                                        return res.redirect(
+                                                            '/check-document-eligibility'
+                                                        );
```
superafroman, line 315: I'd be pretty tempted to throw in another else here and throw an error - currently it's a black whole if appType isn't one of 2, 3, or 4 - this probably never happens but good to code defensively. (Of course throwing an Error here might crash the app so need to handle it properly, maybe just a log statement....)

jenbutongit: I find this pattern slightly cleaner than using multiple else/ifs or a switch statement. Especially since you're in callback hell and everything is so nested, it gets hard to read. This requires no ifs 🤓 

```suggestion
const appTypeRedirect = {
  2: res.redirect( '/business-document-quantity?pk_campaign=DropOff-Service&pk_kwd=DropOff' ),
  3: res.redirect( '/business-document-quantity?pk_campaign=DropOff-Service&pk_kwd=DropOff'),
  4: res.redirect('/check-document-eligibility' )
}
const redirect = appTypeRedirect[req.session.appType]

return redirect
```

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -14,20 +17,30 @@ const CheckUploadedDocumentsController = {
...
             });
-            res.redirect('/check-uploaded-documents');
+
+            CheckUploadedDocumentsController._checkDocumentCountAndPaymentDetails(
+                req,
+                res
+            );
         } catch (err) {
-            sails.log.error(err);
+            sails.log.error(err.message, err.stack);
```
superafroman: Any reason this is different to above? `sails.log.error(err);`?

jenbutongit: Will this catch one error, and then stop executing the loop? or will it iterate through each of them, even though one failed?

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -38,11 +51,149 @@ const CheckUploadedDocumentsController = {
             ? uploadedFile.storageName
             : uploadedFile.location;
 
+        if (!sessionData.appId) {
+            throw new Error('Missing applicaiton id');
```
jenbutongit: sp `application`

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -38,11 +51,149 @@ const CheckUploadedDocumentsController = {
...
+        }
         return {
-            application_id: sessionData.appId || 0, // TODO throw error if this value is false
+            application_id: sessionData.appId,
             uploaded_url: fileUrl,
         };
     },
+
+    _checkDocumentCountAndPaymentDetails(req, res) {
```
jenbutongit, line 25: the `_` nomenclature was a ye olde way of marking something as "private" in js. `_checkDocumentCountAndPaymentDetails` is being called from outside of this class, so you may as well drop the `_` (same with the other methods in this class)

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -38,11 +51,149 @@ const CheckUploadedDocumentsController = {
...
+        };
+        CheckUploadedDocumentsController._checkDocumentCountInDB(
+            documentCountParams,
+            res
+        );
+        // Redirect to payment handled here
+        CheckUploadedDocumentsController._checkPaymentDetailsExistsInDB(
+            paymentParams,
+            res
+        );
```
superafroman, line 44: Because these calls are async I'd chain them, or at least in some way make sure that they're not running in parallel, if soemthing goes wrong in `_checkDocumentCountInDB` we still call `_checkPaymentDetailsExistsInDB`

docs/e-app-submission.md
```diff
@@ -0,0 +1,5 @@
+Uploaded files are not put into the DB until the user clicks 'Continue' on the check your documents/application page because by this stage they've done all their uploading and deleting and have made up their mind, (I'm assuming).
+
+Once they hit continue here, they're happy with what they have and are happy to proceed to payment.
+
+Risks: If a user hits continue here and then goes back (somehow) to delete documents, the doucment they delete is NOT removed from the database.
```
jenbutongit, line 5: I don't have access to your jira, so I can't see the acceptance criteria. It's worth chatting to your lead dev on this and possibly product manager so they know these risks if they don't already.

tests/specs/controllers/CheckUploadedDocuments.test.js
```diff
@@ -0,0 +1,376 @@
...
+        it('should check document count and payment details', () => {
+            // when
+            sandbox.spy(
+                CheckUploadedDocumentsController,
+                '_checkDocumentCountInDB'
+            );
+            sandbox.spy(
+                CheckUploadedDocumentsController,
+                '_checkPaymentDetailsExistsInDB'
+            );
```
jenbutongit: I would do this, so then your calls further down are a bit cleaner/readable
```suggestion
            const checkCountSpy = sandbox.spy(
                CheckUploadedDocumentsController,
                '_checkDocumentCountInDB'
            );
            const checkPaymentSpy = sandbox.spy(
                CheckUploadedDocumentsController,
                '_checkPaymentDetailsExistsInDB'
            );
            //... 
            expect(checkCountSpy.calledWith(...)
```

tests/specs/controllers/CheckUploadedDocuments.test.js
```diff
@@ -0,0 +1,376 @@
...
+            const updateApplicationPaymentDetails = sandbox.stub(
+                ApplicationPaymentDetails,
+                'update'
+            );
+            createUploadedDocumentsUrls.resolves();
+            findUserDocumentCount.resolves(true);
+            updateUserDocumentCount.resolves();
+            findApplicationPaymentDetails.resolves(true);
+            updateApplicationPaymentDetails.resolves();
```
jenbutongit: double check this, but I'm pretty sure you don't need to declare a spy (and assign it) to be able to mock out what it resolves.

I _think_ you can do something like 
```
sandbox.stub(UploadedDocumentUrls, 'create').resolves()
```
this is better since it doesn't look like you are using these spies anyway, just stubbing them.

jenbutongit: also, if you don't know already, it's worth reading up on dependency injection / inversion of control and how it makes testing easier in OOP. It's probably not right for this project now, since it could be a big refactor which we don't have time for

views/eApostilles/applicationSubmissionSuccessful.ejs
```diff
@@ -0,0 +1,5 @@
+<h1 class="heading-xlarge">Application successfully submitted</h1>
+
+<section class="column-two-thirds">
+
+</section>
```
jenbutongit: is this required?

jenbutongit (changes requested):

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Lol, I personally like the '_' before function name to mark it as private. If it were typescript I'd use the 'private' keyword, but I find it helpful to know which functions are public and private at a glance.

The `_checkDocumentCountAndPaymentDetails ` function shouldn't be being called outside the class that's odd 🤔  I need to change that. Where did you see it?

views/eApostilles/applicationSubmissionSuccessful.ejs
RichardBray: Good question, so this is just to have something to show when the payment is successful. It will be fleshed out in the future ticket but this is just a placeholder. So not required for this feature but helpful for the overall flow.

api/controllers/CheckUploadedDocumentsController.js
RichardBray: So the one above prints out the whole error stack, but in one of the unit tests I wanted to test just the message and not the whole stack, so I amended it to split the message and the stack so I can at least test the message is printed.

RichardBray: Hmmm, I was under the assumption when you throw an Error in a try/catch without a 'finally' it will break the loop by default. Is this not the case 😱

superafroman: An error will break the loop, but this promise won't throw an error synchronously so all files will be processed at the same time and an error in the promise will be unhandled. (I've read that back and it's confusing so if you don't get it we can chat through)

RichardBray: Ah I see, yes that does make sense. Okay, I'll look through the code and try and resolve it.

- [x] Look into this

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Yes, that's a good point. I will do that.

- [x] Chain promises to make sure one happens before the other

api/controllers/CheckUploadedDocumentsController.js
jenbutongit: oops, I didn't! (hard to tell in GitHub view).. ignore me :) 

`#` is available in node 14 (instead of `_`), but looks like you're running 12

views/eApostilles/applicationSubmissionSuccessful.ejs
jenbutongit: worth commenting out or deleting entirely. take a look at the "YAGNI" principle 👀

docs/e-app-submission.md
RichardBray: Good point. I actually need to update this documentation as a few things have changed but I'll put the risks in the ticket and mention it to Dan.

We might make some tickets to fix it later and put it in the backlog.

E3sSk_EXIAMwCh5

jenbutongit: is there an error that gets logged when this happens? you could possibly monitor how many times a user actually does it before it requires fixing :)

views/eApostilles/applicationSubmissionSuccessful.ejs
RichardBray: I'm learning a lot from you on this PR, "YAGNI" principle 🤦  The things that people come up with.

RichardBray: but aside from that, you make a good point :)

api/controllers/CheckUploadedDocumentsController.js
jenbutongit: it might be better as a `trace` or `debug`, so you don't get both errors in the logging.

views/eApostilles/applicationSubmissionSuccessful.ejs
jenbutongit: even though it's a handful of KB, it could be sitting collecting dust until the feature is prioritised, and you wont know when.  If another developer picks it up, they might think it was put there for a reason, and they'd create another <section/> just in case that empty <section/> breaks something/is required for something.

api/controllers/CheckUploadedDocumentsController.js
jenbutongit: (feel free to resolve this comment)

api/controllers/CheckUploadedDocumentsController.js
superafroman: Have thought of another thing to consider on this - what you want the behavious to be if `_createDocumentCountInDB` but `_createPaymentDetailsInDB` fails. Currently the doc count record would be in the database and the payment details record wouldn't be. From the looks of the code this probably doesn't matter so much, user will hopefully resubmit and the second time around the count record will be updated rather than created and all will be well in the world.

If it does matter though - there may be other areas of the app where having one record without the other will stuff things up. Or maybe just to maintain data integrity ou don't want incomplete submissions. You can wrap the sequalize calls in a transaction so if any fail you rollback all and keep the DB clean for the next run.

api/controllers/ApplicationTypeController.js
RichardBray: Yeah, that's a fair point. I'll make a note to update this.

- [x] Add else for if appType is not 1,2,3 or 4

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Ah okay, I'll give that a go

- [ ] Change error type here to trace or debug

api/controllers/ApplicationController.js
RichardBray: Yeah, I guess it's just me being lazy but there were already console logs in place and I didn't want to change what they had too much but it makes sense.

- [x] Use sails logger here

api/controllers/ApplicationTypeController.js
RichardBray: Totally agree, again I was just adding to what they had already but it makes sense to refactor it

- [x] Use object here instead of long if/else

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Ha no worries, one less thing to do 💪

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Ah yeah, this is a good point. I like the idea of chaining the whole thing in promises but I don't want to affect `renderPage`  🤔  I will look into it.

- [x] Execute `_checkDocumentCountAndPaymentDetails ` only after forEach is complete

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Interesting, that's something I don't have a lot of experience with, Ali mentioned it to me today actually. I'll take a look at that too.

- [ ] Look into transactions to keep data clean

tests/specs/controllers/CheckUploadedDocuments.test.js
RichardBray: Ah thanks for that, I've read about DI before but never really looked into how it would help tests. Unfortunately, you're right about it being a refactor we don't have time for but definitely something to keep in mind for a greenfield project.

docs/e-app-submission.md
RichardBray: Ticket added to the backlog. Will flag this in standup too.

Screenshot 2021-07-28 at 18 13 38

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Hmm, when I first read your comment I thought you were referring to the sails logging options. https://sailsjs.com/documentation/concepts/logging 

But it doesn't have a `trace`. Which has got me thinking maybe you were referring to something else 🤔

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Actually, now that I'm looking at the code again I think it's not that big of a deal. As in, the document count can be created or updated without the payment details and vice versa. However, both bits of information are needed before the user can pay it doesn't really matter what order they go in.

So technically the user would try again, and eventually, call if they can't submit. Something to test in the MVP I guess 🤷‍♂️

api/controllers/CheckUploadedDocumentsController.js
jenbutongit: you can go for `debug` or `info`. A lot of loggers will provide `trace` so that's where I got it from. trace is usually more verbose than debug

api/controllers/ApplicationController.js
```diff
@@ -216,36 +216,40 @@ var applicationController = {
...
             if (application !== null) {
-              console.log(id + " - has returned from barclays");
+              sails.log.info(id + " - has returned from barclays");
               if (application.submitted === 'draft') {
-                console.log(id + " - has not been submitted previously");
-                console.log(id + " - exporting app data");
-                applicationController.exportAppData(req, res);
+                sails.log.info(id + " - has not been submitted previously");
+                sails.log.info(id + " - exporting app data");
+                applicationController.exportAppData(req, application);
```
jenbutongit: I think the "exporting app data" log should be in ApplicationController.exportAppData - since that is what is `doing` the exporting. same goes with some other method calls/logs you've got

api/controllers/ApplicationController.js
```diff
@@ -216,36 +216,40 @@ var applicationController = {
...
         }).then(function(application) {
             if (application !== null) {
-              console.log(id + " - has returned from barclays");
+              sails.log.info(id + " - has returned from barclays");
               if (application.submitted === 'draft') {
-                console.log(id + " - has not been submitted previously");
-                console.log(id + " - exporting app data");
-                applicationController.exportAppData(req, res);
+                sails.log.info(id + " - has not been submitted previously");
+                sails.log.info(id + " - exporting app data");
```
jenbutongit: sails is using captains-log for their logger https://github.com/balderdashy/captains-log, so there's some more useful documentation there. 

I think we can clean up these lines by using multiple arguments (it looks like captains-log/sails.log will accept multiple). It means there wont be a need to do `+ "string"` as well.
`sails.log.info(id, "has not been submitted previously", "exporting app data")`

api/controllers/ApplicationController.js
```diff
@@ -216,36 +216,40 @@ var applicationController = {
...
     submitApplication: function(req, res) {
         var id = req.query.merchantReturnData;
-        console.log(id + " - attempting to submit application");
+        sails.log.info(id + " - attempting to submit application");
         Application.findOne({
             where: {
                 application_id: id
             }
         }).then(function(application) {
             if (application !== null) {
```
jenbutongit: there are ifs inside ifs here - as well s inside a callback which makes it hard to read. What is returned when `application` is null? 

I use "guard" clauses (you may have seen in other languages or TS). Something like this
```.js
if (application === null) {
  return "whatever"
}
// it is safe to carry on without an else, because it's already done a "return". It wont execute any further
if (application.submitted) {
//...
} else {
//...
}

if (application.serviceType == 1) {
//..
} else if (....) {
//..
}

```

api/controllers/ApplicationController.js
```diff
@@ -216,36 +216,40 @@ var applicationController = {
...
               }
             } else {
-                console.log(id + " - could not be found in db");
+                sails.log.info(id + " - could not be found in db");
             }
         }).catch(function(error) {
-          console.log(id + " - has encountered an error");
-          console.log(id + error);
+          sails.log.info(id + " - has encountered an error");
+          sails.log.info(id + error);
```
jenbutongit: this should be something like `sails.log.error(id, "has encountered an error", error)`. 

What usually happens in production is that these logs are filtered by level. So, in dev envs or test envs you might want to see more logs, but maybe not "debug" logs. You don't want ALL the logs in production because they have to be stored somewhere - so filling them up with lower level logs (when you know it's probably fine) can get costly (and clogs up the logs so it's difficult to see the errors!)

api/controllers/ApplicationController.js
```diff
@@ -355,7 +359,7 @@ var applicationController = {
                     return null;
                   }).catch(function (error) {
                   sails.log(error);
-                  console.log(error);
+                  sails.log.info(error);
```
jenbutongit: just one of these is needed `sails.log.error(error)`. I suggested using `trace` because you were using the stack trace to debug, but there's no need to print it twice if you're not using it to debug anymore (if that makes sense?)

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -14,20 +17,32 @@ const CheckUploadedDocumentsController = {
             if (uploadedFileData.length === 0) {
                 throw new Error('No files uploaded');
             }
-            uploadedFileData.forEach(async (uploadedFile) => {
-                await UploadedDocumentUrls.create(
-                    CheckUploadedDocumentsController._dbColumnData(
-                        uploadedFile,
-                        req
+            for (let i = 0; i <= uploadedFileData.length; i++) {
+                const endOfLoop = i === uploadedFileData.length;
```
jenbutongit: this is minor, something to keep in mind for next time, so no need to change for now. 

`.length` is base 1 (not base 0) - so your `endOfLoop` if statement will occur after all of the files have been iterated through, if that makes sense? just double check that's all working as you intended!

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -14,20 +17,32 @@ const CheckUploadedDocumentsController = {
...
+                if (endOfLoop) {
+                    CheckUploadedDocumentsController._checkDocumentCountAndPaymentDetails(
+                        req,
+                        res
+                    );
+                } else {
```
jenbutongit: (This is minor, so no need to action) 

A guard clause can be used here like I mentioned above to unnest some of this logic. 
```suggestion
                if (endOfLoop) {
                    CheckUploadedDocumentsController._checkDocumentCountAndPaymentDetails(
                        req,
                        res
                    );
                    break //break keyword will "complete" the loop (exit it, like a return) 
                } 
```

jenbutongit (changes requested):

api/controllers/CheckUploadedDocumentsController.js
RichardBray: Yeah, it's a bit confusing but the idea is that after the data in the loop is finished, as in, all the data has been processed, then run the `_checkDocumentCountAndPaymentDetails ` function.

I guess I could also run it on `ploadedFileData.length -1` as well 🤔

api/controllers/ApplicationController.js
RichardBray: Ha, I've actually used these a lot but didn't know they were called guard clauses. You learn something new everyday :)

api/controllers/ApplicationController.js
```diff
@@ -216,36 +216,41 @@ var applicationController = {
...
-                console.log(id + " - exporting app data");
-                applicationController.exportAppData(req, res);
-              } else {
-                console.log(id + " - has been submitted previously");
-              }
-              if (application.serviceType == 1) {
-                console.log(id + " - displaying standard confirmation page to user");
+            if (application === null) {
+                sails.log.info(id + ' - could not be found in db');
+            }
```
jenbutongit:
```suggestion
            if (application === null) {
                sails.log.info(id + ' - could not be found in db');
                return
            }
```
this block needs to return, otherwise the rest of the code will continue executing (unless you want it to keep executing!)

api/controllers/ApplicationController.js
```diff
@@ -436,31 +440,36 @@ var applicationController = {
...
+            })
+                .then(function () {
+                sails.log.info('queued ' + appId);
+                })
+                .catch(Sequelize.ValidationError, function (error) {
+                sails.log.error(error);
+                });
+            })
+            .catch(function (error) {
+            sails.log.error(error);
```
jenbutongit: nitpicking, is the indentation right here? have you run prettier against it? (or was that another PR)

jenbutongit (changes requested):

api/controllers/ApplicationController.js
RichardBray: I'm a bit hesitant to run prettier on this file since most of the code wasn't written by me and wasn't originally formatted correctly, so this PR would have been full of formatting changes making it difficult to see what has actually been changed.

But yeah it's a good spot, I hadn't noticed it before 👍

api/controllers/ApplicationController.js
RichardBray: Good spot, didn't know why I didn't pick up on this before 🤦

jenbutongit (approved):

jenbutongit:
LGTM @RichardBray, left the code in a better state than you found it 💪🤓. 

just some references on the comments I made if you want to do any further reading:
- condition objects - https://betterprogramming.pub/dont-use-if-else-and-switch-in-javascript-use-object-literals-c54578566ba0
- js "guard" clauses - https://blog.webdevsimplified.com/2020-01/guard-clauses/

PR #256, opened 2021-07-30
Update wording to turnaround times / email address

benjamindi96k:


views/applicationForms/applicationSubmissionSuccessful.ejs
```diff
@@ -136,7 +136,7 @@ var fastTrackApplication = false
     <h2 class="heading-medium">How long it takes and when you'll be charged</h2>
 
 
-    <p>It normally takes 2 working days to process your documents, plus courier or postage time. It might take longer during Easter and summer.</p>
+    <p>Due to high demand applications are currently taking up to 10 working days, plus courier or postage time. It might take longer during Easter and summer.</p>
```
simonmoore100: I would personally drop the bit about taking longer in Easter or summer. We are in summer now and updating the text so I think its fine to drop it. We can revert to this text when the office gets back to normal.

simonmoore100 (changes requested):
Just remove that one sentence and its good to go

simonmoore100 (approved):
LGTM

PR #257, opened 2021-08-03
chore: add documents to the db only after successful payment

RichardBray:


api/controllers/EAppSubmittedController.js
```diff
@@ -0,0 +1,59 @@
...
+const EAppSubmittedController = {
+    addDocsAndRenderPage(req, res) {
+        try {
+            const { uploadedFileData } = req.session.eApp;
+            if (uploadedFileData.length === 0) {
+                throw new Error('No files uploaded');
+            }
+            for (let i = 0; i <= uploadedFileData.length; i++) {
+                const endOfLoop = i === uploadedFileData.length;
+                if (endOfLoop) {
```
superafroman, line 14: You could just do this outsde the loop?

api/controllers/EAppSubmittedController.js
```diff
@@ -0,0 +1,59 @@
...
+                    )
+                )
+                    .then(() => {
+                        sails.log.info(
+                            `Url for document ${uploadedFileData[i].filename} added to db`
+                        );
+                    })
+                    .catch((err) => {
+                        throw new Error(err);
+                    });
```
superafroman, line 30: This call is async and may not complete inside the loop or before `EAppSubmittedController._renderPage(res)` has been called. You either need to `await` or put all promises in an array and do a Promise.all to make sure everything is complete before rendering success.

superafroman: Another thing to consider... DB transactions - if one of these promises fails, the data won't be inserted, so should the user try again? And if so will they be submitting all their data again? We may want to rollback other successful inserts if one fails, so that resubmission doesn't create duplicates, or produce an error if there's a primary key based on the dbColumnData.

tests/specs/controllers/CheckUploadedDocuments.test.js
```diff
@@ -2,7 +2,7 @@ const { expect } = require('chai');
 const sinon = require('sinon');
 const CheckUploadedDocumentsController = require('../../../api/controllers/CheckUploadedDocumentsController');
 
-describe('CheckUploadedDocumentsController', () => {
+describe.only('CheckUploadedDocumentsController', () => {
```
superafroman, line 5: Should this have been merged?

superafroman (commented):
Appreciate this is already merged but thought I'd leave a couple of comments anyway :).

PR #260, opened 2021-08-06
Feature/eia 101

RichardBray:
# Description

 **1 - FEATURE VIDEO RUN-THROUGH 🎨**

https://user-images.githubusercontent.com/1377253/128555209-8319cb1f-8df2-4491-a0a8-415a9f9f6d43.mp4

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-101

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:

### DIFF
--- a/api/controllers/DashboardController.js
+++ b/api/controllers/DashboardController.js
@@ -7,7 +7,6 @@ var summaryController = require('./SummaryController');
 
 
 var dashboardController = {
-
     /**
      * Move all relevent Application data provided by the user into the Exports table.
      * This table can then be exported as a JSON object directly to the Submission API.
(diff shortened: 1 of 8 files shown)


database_change_log/EIA-101_update_dashboard_data_function.sql
```diff
@@ -0,0 +1,163 @@
...
+	LIMIT $2 OFFSET $3'
+USING _user_id, _limit, _offset;
+ END IF;
+END;
+$BODY$
+  LANGUAGE plpgsql VOLATILE
+  COST 100
+  ROWS 1000;
+ALTER FUNCTION public.dashboard_data_eapp(integer, integer, integer, text, text, text, text, text)
+  OWNER TO postgres;
```
PaulioRandall: It think it will be well worth formatting this SQL a bit more for readability. Mostly, indention of the string literal queries. I avoid rigidly following a specific style but I usually go with something like the example below. It keeps lines short and avoids too much nesting. **Any style focusing on clarity, readability, and easy maintenance is good in my book.**
```
SELECT
    table_1.column_a,
    table_1.column_b,
    table_2.column_c
FROM
    table_1
INNER JOIN
    table_2
    ON table_1.column_a = table_2.column_a
WHERE
    table_1.column_b = 'criteria'
    AND table_2.column_c = 'criteria'
```

RichardBray: Good shout, I'm not much of an expert with sql files so I didn't know if there were any formatting standards, but I'll look around for a VS Code plugin. I'm sure one exists.

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+                        pageSize,
+                        offset,
                         sortOrder: Math.abs(sortOrder).toString(),
-                        direction: direction,
+                        direction,
                         queryString: '%' + searchCriteria + '%',
-                        secondarySortOrder: secondarySortOrder,
-                        secondaryDirection: secondaryDirection
+                        secondarySortOrder,
+                        secondaryDirection,
```
jenbutongit, line 66: modern ES 💪

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+            offset,
+            sortOrder,
+            currentPage,
+            searchCriteria,
+            pageSize,
+            req,
+            res,
+        } = displayAppsArgs;
+        //redirect to 404 if user has manually set a page in the query string
+        var resultCount = 0;
```
jenbutongit: is there a reason why this is set to 0, rather than just `const resultCount = results.length`? results.length is used a few times

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
-                                paginationMessage = 'No applications found';
-                            }
-                            else {
-                                paginationMessage = 'Showing ' + (offset + 1) + ' &ndash; ' + pageUpperLimit + ' of ' + resultCount + ' applications submitted in the last 60 days';
-                            }
+                    var certPath;
+                    try {
+                        certPath = req._sails.config.casebookCertificate;
+                    } catch (err) {
+                        console.error('Null certificate path: [%s] ', err);
```
jenbutongit: should this be a sails log?

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
-
-            //console.log(req.session.passport.user, pageSize, offset,Math.abs(sortOrder).toString(),direction, '%' + searchCriteria + '%', secondarySortOrder, secondaryDirection);
-            sequelize.query(userApplicationsSql,
-                {
+        Application.count({
+            where: { user_id: req.session.passport.user },
+        }).then((totalApplications) => {
+            HelperService.refreshUserData(req, res).then(() => {
+                var pageSize = 20;
+                var currentPage = req.query.page || 1;
```
jenbutongit, line 42: does this start from base 0? or do the pages start from 1?

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
-                            // First build array of application references to be passed to the Casebook Status API for this page. Can submit up to 20 at a time.
-
-                            var applicationReferences = [];
-
-                            for ( var i=0; i < results.length; i++ ) {
-                                applicationReferences.push(results[i].unique_app_id);
-                            }
+                    type: sequelize.QueryTypes.SELECT,
+                };
+                const userData = HelperService.getUserData(req, res);
```
jenbutongit: what happens when you do `getUserData` when there isn't a user logged in? could you move this line up to the top where you check if a user is present?

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+        sequelize
+            .query(queryApplications, storedProcedureArgs)
+            .then((electronitApplications) => {
+                dashboardController._displayApplications(
+                    electronitApplications,
+                    displayAppsArgs
+                );
+            })
+            .catch((error) => {
+                sails.log(error);
```
jenbutongit: this is a silent error - so whatever is calling it wont know if anything failed - is that intended? (or out of scope to deal with right now?_

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
 
-                                if (err) {
-                                    console.log("Casebook Status Retrieval API error: ", err);
-                                } else if (api_results[0].length === 0){
-                                    console.log("No Casebook Statuses available");
-                                } else {
+                function (callback) {
+                    var request = require('request');
+                    var crypto = require('crypto');
+                    var apiQueryString = require('querystring');
```
jenbutongit: what's the reason for doing a nested require?

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+                    // First build array of application references to be passed to the Casebook Status API for this page. Can submit up to 20 at a time.
 
+                    var applicationReferences = [];
 
-                                    for ( var k=0; k < api_results[0].length; k++ ) {
-                                       appRef[ api_results[0][k].applicationReference ] = api_results[0][k].status;
-                                       trackRef[ api_results[0][k].applicationReference ] = api_results[0][k].trackingReference;
+                    for (var i = 0; i < results.length; i++) {
+                        applicationReferences.push(results[i].unique_app_id);
+                    }
```
jenbutongit: it doesn't look like the reason this was a for loop to begin with is still there, so this might be cleaner as a `map` now

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+                                callback(true);
+                                return;
+                            } else if (response.statusCode == 200) {
+                                const obj = body;
+                                callback(false, obj);
+                            } else {
+                                console.log(
+                                    'Invalid response from Casebook Status API call: ',
+                                    response.statusCode
+                                );
```
jenbutongit: this will be more useful as a sails warning imo

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+                        for (let result in results) {
+                            result.app_status = appRef[result.unique_app_id];
+                            result.tracking_ref =
+                                trackRef[result.unique_app_id];
                         }
-                    );
+                    }
+                }
+
+                var moment = require('moment');
```
jenbutongit: moment is self deprecated now.. leave it in there for now but add a ticket to make sure this is replaced

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+
     /**
      * @function openCoverSheet
      * @description Open the cover sheet
      * @param req {Array} - request object
      * @param res {Array} - response object
      * @return summary
      */
-    openCoverSheet: function(req, res) {
+    openCoverSheet: function (req, res) {
```
jenbutongit: this is bizarre (I know it's not your code!).. could this be changed to a normal method? 

```suggestion
    openCoverSheet (req, res) {
```

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+                    res.view('404');
+                }
+            });
         } else {
             res.view('404');
         }
-    }
-
-
+    },
```
jenbutongit: this might be nicer with guard clauses ;). I think really only one `res.view(404)` needs to be used. If summaryController.fetchAll(..) doesn't return anything, the only other thing that should be returned is a 404. 

something like 
```
if(result) {
  // return something
}
res.view(404)
```

api/controllers/DashboardController.js
```diff
@@ -16,275 +14,360 @@ var dashboardController = {
...
+                    //Add tracking reference to results array
+                    // Only update if there are matching values
+
+                    if (err) {
+                        console.log(
+                            'Casebook Status Retrieval API error: ',
+                            err
+                        );
+                    } else if (api_results[0].length === 0) {
+                        console.log('No Casebook Statuses available');
```
jenbutongit: sails log?

tests/specs/controllers/Dashboard.test.js
```diff
@@ -8,35 +8,171 @@
...
+                        },
+                        casebookCertificate: '',
+                        casebookKey: '',
+                        hmacKey: '',
+                    },
+                },
+                params: {
+                    unique_app_id: '123',
+                },
+                allParams: sandbox.stub().returns({
```
jenbutongit: stubs are useful to change the behaviour of a method and/or check how many calls or the arguments that it was passed when it was called. If you aren't using these features you can just do this

`allParams: () => ({dashboardFilter: ''})`

views/eApostilles/dashboard.ejs
```diff
@@ -0,0 +1,44 @@
+<% pageTitle="Your account - Get a document legalised – GOV.UK" %>
+    <div class="div-dashboard">
+        <div class="inner_header <%if (!user_data.loggedIn){%>no-user-signed-in<%}%>">
+            <%- partial ('../partials/inner-header.ejs')%>
+
+        </div>
+
+        <h1 class="heading-xlarge">Your account</h1>
+        <% if (typeof(message)!='undefined' && message.length>0) {
```
jenbutongit: might just be easier to do a double `!`, `if(!!message)`?

views/partials/dashboardResultsEApp.ejs
```diff
@@ -0,0 +1,146 @@
+<div class="container">
+    <form id="dashboard-search-table-form" method="post" action="/dashboard">
+        <label for="dashboard-search-filter" class="visuallyhidden">Search by reference or service</label>
+        <input type="text" name="dashboardFilter" id="dashboard-search-filter" style="height: 40px;padding: 0 5px;"
+            placeholder="Search by reference or service" title="Search" value="<%= searchCriteria %>"
+            class="govuk-input govuk-!-width-one-third">
+        <input class="submit" type="submit" value="Search">
+        &nbsp;
```
jenbutongit: do you need this `&nbsp;`?

views/partials/dashboardResultsEApp.ejs
```diff
@@ -0,0 +1,146 @@
...
+        <% if (searchCriteria) { %><a href="/dashboard" id="dashboard-clear-results">Clear search filter</a>
+            <%}%>
+    </form>
+</div>
+<table id="previousApplications">
+    <caption class="visually-hidden">Application <% if (searchCriteria) { %>search results<%}else{%>history<%}%> table.
+                    Can be sorted by clicking the table headers.</caption>
```
jenbutongit, line 14: I suppose only people with screenreaders would be able to "see" (hear) this. Those users usually aren't "clicking". maybe selecting is a better term?

jenbutongit (changes requested):

api/controllers/DashboardController.js
RichardBray: I believe it starts from 1 but this was already there when I got the code so I'm assuming it works fine 🤷‍♂️

api/controllers/DashboardController.js
RichardBray: fair point, could have a guard clause a bit higher up 🤔

api/controllers/DashboardController.js
RichardBray: yep, good shout

api/controllers/DashboardController.js
RichardBray: 😄

api/controllers/DashboardController.js
RichardBray: no, no reason. I guess it's a way to tell the compiler/whoever is reading this that this variable is going to be of type number, but I'm not to bothered by it.

api/controllers/DashboardController.js
RichardBray: No reason, I saw it too but I had my if-it-aint-broke-dont-fix-it mentality on so I just left it, but yeah it does make sense to have all the imports/requires at the top.

api/controllers/DashboardController.js
RichardBray: I think it's the old school way of doing it, but it still works, although it does look weird

api/controllers/DashboardController.js
RichardBray: Good spot, I'll update this code

api/controllers/DashboardController.js
RichardBray: I'm not sure I follow, how did you figure out it's a silent error? Is it because it's missing a log method?
`sails.log.error(error);` instead of `sails.log(error);`

views/partials/dashboardResultsEApp.ejs
RichardBray: I think so, this is something I took from the dashboardResults.ejs page but yeah it does make sense.

api/controllers/DashboardController.js
RichardBray: Ticket written.
https://jira.digital.csd.fcdo.gov.uk/browse/EIA-151

I'll bring it up in standup tomorrow

api/controllers/DashboardController.js
jenbutongit: silent wasn't the best term sorry 😄, I meant if just logging the error is OK - and does not affect the rest of the flow of the code/there's no need to "recover" from the code you can ignore this. (for example redirecting the user to an error page etc)

RichardBray: gotcha, yeah that makes a lot of sense, I've made the updates now.

jenbutongit (approved):

PR #264, opened 2021-08-12
Feature/eia 99 - Delete temporary documents after timeout

RichardBray:
# Description

 **[1 - FEATURE VIDEO RUN-THROUGH 🎨](https://drive.google.com/file/d/1sfPaf5q53R_OodFxK_cympWLBsZyL6JI/view?usp=sharing)**

https://user-images.githubusercontent.com/1377253/129396588-301aa0e0-c5e4-498f-9567-762ee45359a1.mp4

 **[2 - CODE EXPLANATION 🧑🏿‍💻](https://drive.google.com/file/d/1b_FNGmnU9C59u34YJfP4Is2iP5K-2rwO/view?usp=sharing)**

https://user-images.githubusercontent.com/1377253/129396665-2f9a8d9b-4268-404b-8173-03255c350fa0.mp4

**3 - UPDATED DIRECTION ↩️**

https://user-images.githubusercontent.com/1377253/131165511-4fa040eb-0d56-4802-836a-3fee96fe7fb9.mp4

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-99

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:


superafroman:
@RichardBray all these changes look good, only question I have is what happens if someone just closes their browser after uploading docs?

RichardBray:
> @RichardBray all these changes look good, only question I have is what happens if someone just closes their browser after uploading docs?

That is actually a good question which I hadn't thought of. So currently the documents will stay in S3. I think there is currently a 7 day retention period for files in S3 but we're planning to reduce that.

superafroman:
> That is actually a good question which I hadn't thought of. So currently the documents will stay in S3. I think there is currently a 7 day retention period for files in S3 but we're planning to reduce that.

I think this is going to be the most common situation (gut feel, no data to back it up..). So if we're OK that the files will live as long as S3 is configured to keep them, why bother clearing them out ourselves at all?

RichardBray:
> > That is actually a good question which I hadn't thought of. So currently the documents will stay in S3. I think there is currently a 7 day retention period for files in S3 but we're planning to reduce that.
> 
> I think this is going to be the most common situation (gut feel, no data to back it up..). So if we're OK that the files will live as long as S3 is configured to keep them, why bother clearing them out ourselves at all?

Good point. I will have a chat with Dan about it

RichardBray:
Hey @superafroman Dan agrees with your point and we're trying to think of the best way to solve the issue. We're thinking of having a different tag for documents that have a submitted application and documents that don't and have different retention periods for both. So the documents that don't have a submitted application could have a shorter retention period, hours instead of days. 

I've done some research and it doesn't look like there's a way to have different retention periods based on tags for files, it seems it can be done for buckets only. But it would be good to get your thoughts or if you want to have a quick call about it. Don't know how much sens I'm making :)

superafroman (approved):
Looking good! 👍

RichardBray:
> Looking good! 👍

Super cool, thanks for reviewing this so quickly 👍

PR #265, opened 2021-08-16
Feature/eia 102 - View in-progress eApostille application

RichardBray:
# Description

 **[1 - FEATURE VIDEO RUN-THROUGH 🎨](https://drive.google.com/file/d/1Jn1D4rQjKSyPY48Cd3GSCZKZbeEzVH31/view?usp=sharing)**

https://user-images.githubusercontent.com/1377253/129601229-fe611cfa-4e31-489e-a197-68949d894f73.mp4

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-102

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:


api/controllers/OpenEAppController.js
```diff
@@ -0,0 +1,144 @@
...
+        }
+    },
+
+    _getApplicationDataFromCasebook(req, res) {
+        const { hmacKey, customURLs, casebookCertificate, casebookKey } =
+            req._sails.config;
+        const queryParamsObj = {
+            timestamp: new Date().getTime().toString(),
+            applicationReference: req.params.unique_app_id,
+        };
```
jenbutongit, line 44: `.getTime()` returns milliseconds since epoch, but some other systems will use seconds since epoch. Just double check which you need (if it matters)

assets/styles/informed.scss
```diff
@@ -2032,3 +2032,45 @@ govuk-checkboxes__item {
...
+  }
+
+  .govuk-table__header {
+    width: 200px;
+  }
+}
+
+.no-table-header-border .govuk-table__head .govuk-table__header {
+  border-bottom: none;
+}
```
jenbutongit: with BEM all rules should all be the same level of specificity (no nesting), although it doesn't seem like the rest of the code is really following BEM. Essentially this could be a headache in the future if you want a table with no borders, but if we need `govuk-table__cell` to have a border  in the future (weird example, I don't know why anyone would want this..) it will be a bit of a headache to add that style since you'd have to override by adding more specificity. 

To add a modifier in the same amount of lines (I'm pretty sure this works 🤔) 
```
.govuk-table__header,
.govuk-table__cell,
.govuk-table__row {
  &--no-border {
    border-bottom: none;
  }
}
```

assets/styles/informed.scss
```diff
@@ -2032,3 +2032,45 @@ govuk-checkboxes__item {
...
+.right-align-last-cell {
+  .govuk-table__row .govuk-table__cell:last-child {
+    text-align: right;
+  }
+}
```
jenbutongit: could this be BEM'd?

views/eApostilles/openEApp.ejs
```diff
@@ -0,0 +1,80 @@
...
+        We will notify you by email when it has been legalised. Your successfully legalised document will then
+        be
+        available for
+        download from this page.
+    </p>
+    <p>
+        You will have 21 days from when the documents have been successfully legalised to download them before
+        they are
+        deleted
+        for security reasons.
```
jenbutongit: what's up with the line breaks here? can it be just one or two lines? it's harder to read when it's all separated.  Is this using the application complete pattern too? https://design-system.service.gov.uk/patterns/confirmation-pages/

views/eApostilles/openEApp.ejs
```diff
@@ -0,0 +1,80 @@
...
+    <table class="govuk-table right-align-last-cell ">
+        <caption class="heading-medium">PDF files</caption>
+        <tbody class="govuk-table__body">
+            <% for (const document of documents) {%>
+            <tr class="govuk-table__row">
+                <td class="govuk-table__cell"><%= document.name %></td>
+                <td class="govuk-table__cell">
+                    <strong class="govuk-tag govuk-tag--blue">
+                        <%= document.status %>
+                    </strong>
```
jenbutongit: one for the design team, maybe BA/product knowledge, but what are all of the statuses? is it OK that they are all blue, even though they might be different?

views/eApostilles/openEApp.ejs
```diff
@@ -0,0 +1,80 @@
...
+                    <th scope="row" class="govuk-table__header">Service</th>
+                    <td class="govuk-table__cell">e-Apostille</td>
+                </tr>
+                <tr>
+                    <th scope="row" class="govuk-table__header">Date submitted</th>
+                    <td class="govuk-table__cell"><%= dateSubmitted %></td>
+                </tr>
+            </tbody>
+        </table>
+    </div>
```
jenbutongit: I just had a peek at the video, you might be better off using this https://design-system.service.gov.uk/components/summary-list/#summary-list-without-borders since you're not iterating through rows here?

views/eApostilles/openEApp.ejs
```diff
@@ -0,0 +1,80 @@
...
+                    <th scope="row" class="govuk-table__header">Original cost</th>
+                    <td class="govuk-table__cell"><%= originalCost %></td>
+                </tr>
+                <tr>
+                    <th scope="row" class="govuk-table__header">Payment reference</th>
+                    <td class="govuk-table__cell"><%= paymentRef %></td>
+                </tr>
+            </tbody>
+        </table>
+    </div>
```
jenbutongit: see comment for +46 to +60

jenbutongit (changes requested):

api/controllers/OpenEAppController.js
RichardBray: Yep, it does matter, the casebook API needs timestamp in milliseconds. When you say other systems do you mean different settings for containerised environments

assets/styles/informed.scss
RichardBray: Ah cool, yeah I haven't really been following BEM with all my new styles. I'll see what I can do to fix this

views/eApostilles/openEApp.ejs
RichardBray: Ha, I've honestly never noticed this. I copied and pasted it from Figma (I think), and this is how it came out. Good spot though.

views/eApostilles/openEApp.ejs
RichardBray: So I originally thought there were only meant to be two types of blues but looks like red needs to be used as well. But those changes are to be done in a separate ticket https://jira.digital.csd.fcdo.gov.uk/browse/EIA-145 so this will all change at some point :)

views/eApostilles/openEApp.ejs
RichardBray: Oh wow, didn't know this existed, ha, if I did I wouldn't have rolled out my own 😆 
Yeah this makes much more sense.

api/controllers/OpenEAppController.js
jenbutongit: I mean all of the services/microservices that are used to make the eApostille service work. Unix/epoch time is usually recorded in seconds. So for other systems you might find that timestamp is stored in seconds, so you would have to do some work to multiply or divide by 1000 when you get the response from the database or server

RichardBray: That's a really good point. To be fair, this is how the project has always made API calls to Casebook so to my knowledge, it's always only going to be milliseconds but I'll double-check with Iizuka.

api/controllers/DashboardController.js
```diff
@@ -108,7 +107,6 @@ var dashboardController = {
             })
             .catch((error) => {
                 sails.log.error(error);
-                res.serverError();
```
jenbutongit, line 111: I assume sails will still handle rendering the error page?

jenbutongit (changes requested):

api/controllers/DashboardController.js
RichardBray: Ah good point, I honestly don't know why I removed that line in the first place

jenbutongit (approved):
lgtm!

PR #266, opened 2021-08-17
Feature/eia 97 - Success page + notification email

RichardBray:
# Description

 **[1 - FEATURE VIDEO RUN-THROUGH 🎨](https://drive.google.com/file/d/1jWB1bolly6_xbhCMei80eBPltHyz5JZi/view?usp=sharing)**

https://user-images.githubusercontent.com/1377253/129744168-ccbc5f7f-ebdc-42e5-9540-a3608adfc187.mp4

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-97

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:

### DIFF
--- a/api/controllers/EAppSubmittedController.js
+++ b/api/controllers/EAppSubmittedController.js
@@ -14,7 +14,8 @@ const EAppSubmittedController = {
             for (let i = 0; i <= uploadedFileData.length; i++) {
                 const endOfLoop = i === uploadedFileData.length;
                 if (endOfLoop) {
-                    return EAppSubmittedController._renderPage(res);
+                    EAppSubmittedController._sendConfirmationEmail();
+                    return EAppSubmittedController._renderPage(req, res);
                 }
                 UploadedDocumentUrls.create(
                     await EAppSubmittedController._dbColumnData(
@@ -37,9 +38,32 @@ const EAppSubmittedController = {
         }
     },
 
-    _renderPage(res) {
-        return res.view('eApostilles/applicationSubmissionSuccessful.ejs', {});
+    _sendConfirmationEmail() {
+        const emailAddress = '';
+        const applicationRef = '';
+        const sendInformation = {firstName: '', lastName: ''};
+        const userRef = '';
+        const serviceType = 4;
+
+        EmailService.submissionConfirmation(
+            emailAddress,
+            applicationRef,
+            sendInformation,
+            userRef,
+            serviceType
+        );
     },
+
+    _renderPage(req, res) {
+        const userData = HelperService.getUserData(req, res);
+        const queryParams = req.params.all();
+
+        return res.view('eApostilles/applicationSubmissionSuccessful.ejs', {
+            user_data: userData,
+            applicationId: queryParams.merchantReference,
+        });
+    },
+
     /**
      * @return {Promise<{application_id: number, uploaded_url: string, filename: string}>}
      **/
--- a/api/controllers/EAppSubmittedController.js
+++ b/api/controllers/EAppSubmittedController.js
@@ -14,7 +14,7 @@ const EAppSubmittedController = {
             for (let i = 0; i <= uploadedFileData.length; i++) {
                 const endOfLoop = i === uploadedFileData.length;
                 if (endOfLoop) {
-                    return EAppSubmittedController._renderPage(req, res);
+                    return EAppSubmittedController._renderPageAndSendConfirmationEmail(req, res);
                 }
                 UploadedDocumentUrls.create(
                     await EAppSubmittedController._dbColumnData(
@@ -37,7 +37,7 @@ const EAppSubmittedController = {
         }
     },
 
-    _renderPage(req, res) {
+    _renderPageAndSendConfirmationEmail(req, res) {
         const queryParams = req.params.all();
         const applicationId = queryParams.merchantReference;
         const userDetails = {
@@ -53,7 +53,6 @@ const EAppSubmittedController = {
             userDetails,
             applicationId
         );
-
         return res.view('eApostilles/applicationSubmissionSuccessful.ejs', {
             loggedIn: userObjectExists,
             email: userDetails.email,
--- a/api/controllers/EAppSubmittedController.js
+++ b/api/controllers/EAppSubmittedController.js
@@ -47,14 +47,12 @@ const EAppSubmittedController = {
             appType: req.session.appType,
             userRef: req.session.user.id,
         };
-        const userObjectExists = req.session.hasOwnProperty('user');
 
         EAppSubmittedController._sendConfirmationEmail(
             userDetails,
             applicationId
         );
         return res.view('eApostilles/applicationSubmissionSuccessful.ejs', {
-            loggedIn: userObjectExists,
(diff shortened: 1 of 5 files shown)


api/services/EmailService.js
```diff
@@ -1,50 +1,54 @@
...
-    {
-      'cache-control': 'no-cache',
-      'content-type': 'application/json'
-    },
-    method: 'POST',
-    json: true,
-    body: postData
-  };
-  return options;
+function setOptions(postData, url) {
```
jenbutongit: it looks like url is unchanged (it's always `/confirm-submission`?), so it might be better to remove this param from the `setOptions` function. 

you might also want to look at this https://www.npmjs.com/package/request#convenience-methods. It looks like you can set a default request, so you don't have to use a function to provide your defaults 🤓 

(also as an aside, request is deprecated, I'm pretty sure node will provide the same functionality `http.request`)

views/eApostilles/applicationSubmissionSuccessful.ejs
```diff
@@ -1 +1,41 @@
...
+<% pageTitle="e-Apostille application submitted" %>
+
+<div class="inner_header no-user-signed-in">
+    <%- partial ('../partials/inner-header.ejs')%>
+</div>
+
+<section class="column-two-thirds">
+    <div class="container-confirmed govuk-!-margin-bottom-5">
+        <h1 class="heading-xlarge confirmed">Application submitted</h1>
+        <div class="heading-large govuk-!-margin-top-0 govuk-!-margin-bottom-0" style="font-weight: normal">
```
jenbutongit: can we provide the inline style via css?

api/services/EmailService.js
jenbutongit: oh oops, yes, it looks like you do use setOptions with two different urls. It still might be worth digging into the convenience methods though

api/controllers/EAppSubmittedController.js
```diff
@@ -37,8 +37,45 @@ const EAppSubmittedController = {
...
+        const applicationId = queryParams.merchantReference;
+        const userDetails = {
+            firstName: req.session.account.first_name,
+            lastName: req.session.account.last_name,
+            email: req.session.email,
+            appType: req.session.appType,
+            userRef: req.session.user.id,
+        };
+
+        EAppSubmittedController._sendConfirmationEmail(
```
jenbutongit, line 51: minor - but is there a reason why `EAppSubmittedController.method` is used instead of `this.method`?

api/services/EmailService.js
jenbutongit: oops, just some more context
`request(setOptions(postData, url), (err, res, body)` is (imo) really hard to read/understand. You've got to understand what setOptions does with postData/url, and it feels cluttered.. I can't explain it 😅. 

By using a default request, it would look something like this (just some pseudo code..) 
```
const emailRequest = request({
baseUrl: sails.config.customURLs.notificationServiceURL
headers,
method: 'POST',
json: true
})

//inside your method(s) you would make the request like so: 
emailRequest({url: '/confirm-submission', body: ...}) //the rest of it is defaulted
```

jenbutongit (changes requested):

views/eApostilles/applicationSubmissionSuccessful.ejs
RichardBray: sure thing, I'll try and BEMify it as well 😄 
Then again, for a one-liner utility class like this is it worth using BEM for this 🤔

api/controllers/EAppSubmittedController.js
RichardBray: I guess it's because I haven't instantiated the object so by default all the methods are static (if that makes sense). If I was using a class or a function constructor I think the `this` keyword would make sense.

RichardBray:
> (also as an aside, request is deprecated, I'm pretty sure node will provide the same functionality http.request)

Good point I was thinking about it but it would have been out of scope for this ticket so I'm going to add it to the ticket where I have to change moment.js as well.

RichardBray:
> you might also want to look at this https://www.npmjs.com/package/request#convenience-methods. It looks like you can set a default request, so you don't have to use a function to provide your defaults 🤓

Really nice suggestion, didn't know this was a thing 👍

RichardBray:
> oops, just some more context
> request(setOptions(postData, url), (err, res, body) is (imo) really hard to read/understand. You've got to understand what setOptions does with postData/url, and it feels cluttered.. I can't explain it

Totally agree, I didn't see it when I wrote it but now that you mention it, it is a bit difficult to understand.

views/eApostilles/applicationSubmissionSuccessful.ejs
jenbutongit: you can use one of govuk's utility classes - `govuk-!-font-weight-regular`

RichardBray: Ah didn't know this existed. I'll do that instead, thanks 👍

api/controllers/EAppSubmittedController.js
jenbutongit: does sails do anything special with the controllers? it's probably not worth doing now since it might look odd when compared to the rest of the codebase, but you could unnest all of the methods and export all of the ones that are public, and not export the private ones. 

your file would look something like this
```.js 
// EAppSubmittedController.js
export function async addDocsAndRenderPage(req, res) {...}

function renderPageAndSendConfirmationEmail(req, res) {...}

// anotherFileThatUsesEAppSubmittedController.js
import { * as EAppSubmittedController } from '../somewhere/EAppSubmittedController'  // imports all of the functions that are "public" 

// so you can use it like so:
EAppSubmittedController.addDocsAndRenderPage(..)

// but not this since it's not exported 
EAppSubmittedController.renderPageAndSendConfirmationEmail(..)

```

api/services/EmailService.js
```diff
@@ -1,50 +1,55 @@
...
-        console.log(err);
-      } else {
-        console.log(res.statusCode, body);
-      }
-    });
-  },
-  failedDocuments: function(email,failed_certs){
-    var url = '/failed-documents';
-    var postData= {to: email, failed_certs: failed_certs};
+const emailRequest = request.defaults({
```
jenbutongit, line 6: 💪

tests/specs/controllers/EAppSubmittedController.test.js
```diff
@@ -120,19 +132,63 @@ describe('EAppSubmittedController', () => {
         });
     });
 
-    describe('_renderPage', () => {
-        it('should render submission page', async () => {
-            // when
+    describe.only('_renderPageAndSendConfirmationEmail', () => {
```
jenbutongit: remove .only

jenbutongit (approved):

tests/specs/controllers/EAppSubmittedController.test.js
RichardBray: Ha good spot 👍

PR #273, opened 2021-08-26
Feature/EIA-88 Triage questions

RichardBray:
# Description

 **1 - FEATURE VIDEO RUN-THROUGH 🎨**

https://user-images.githubusercontent.com/1377253/131149081-0864f8ec-5d4e-40f4-9cac-de2873379300.mp4

 **2 - CODE VIDEO RUN-THROUGH 🎨**

https://user-images.githubusercontent.com/1377253/131151778-ddfba1da-ba61-4bc0-bada-093e01ed6ae1.mp4

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-88

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:

### DIFF
--- a/api/controllers/EAppEligibilityQuestionsController.js
+++ b/api/controllers/EAppEligibilityQuestionsController.js
@@ -9,6 +9,12 @@ const EAppEligibilityQuestionsController = {
                 'eApostilles/eligibilityQuestionThree.ejs',
         };
 
+        const paramMatchesViewRoute = Object.keys(eligibilityViews).includes(req.param('question'));
+
+        if (!paramMatchesViewRoute) {
+            return res.view('404.ejs');
+        }
+
         const questionPage = eligibilityViews[req.param('question')];
         const userData = EAppEligibilityQuestionsController._fetchUserData(
             req,
--- a/api/controllers/EAppEligibilityQuestionsController.js
+++ b/api/controllers/EAppEligibilityQuestionsController.js
@@ -23,6 +23,32 @@ const EAppEligibilityQuestionsController = {
         );
     },
 
+    _pageRendering(req, res, pagePath) {
+        const userData = EAppEligibilityQuestionsController._fetchUserData(
+            req,
+            res
+        );
+
+        return res.view(pagePath, {
+            user_data: userData,
+            page_error: false,
+        });
+    },
+
+    _fetchUserData(req, res) {
+        const userData = HelperService.getUserData(req, res);
+        EAppEligibilityQuestionsController._checkUserLoggedIn(userData, res);
+
+        return userData;
+    },
+
+    _checkUserLoggedIn(userData, res) {
+        if (!userData.loggedIn) {
+            sails.log.error('User is not logged in', userData);
+            return res.forbidden();
+        }
+    },
+
     handleQuestionOneAnswer(req, res) {
         const params = {
             radioInputName: 'eapostille-acceptable',
@@ -72,7 +98,7 @@ const EAppEligibilityQuestionsController = {
     },
 
     _handleYesNoAnswers(req, res, params) {
-        const {radioInputName, errorPagePath, redirectOptions} = params;
+        const { radioInputName, errorPagePath, redirectOptions } = params;
         const radioValueSelected = req.body[radioInputName];
         const userData = EAppEligibilityQuestionsController._fetchUserData(
             req,
@@ -88,32 +114,6 @@ const EAppEligibilityQuestionsController = {
         }
 
         return res.redirect(redirectOptions[radioValueSelected]);
-    },
-
-    _pageRendering(req, res, pagePath) {
-        const userData = EAppEligibilityQuestionsController._fetchUserData(
-            req,
-            res
-        );
-
-        return res.view(pagePath, {
-            user_data: userData,
-            page_error: false,
-        });
-    },
-
-    _fetchUserData(req, res) {
-        const userData = HelperService.getUserData(req, res);
-        EAppEligibilityQuestionsController._checkUserLoggedIn(userData, res);
-
-        return userData;
-    },
-
-    _checkUserLoggedIn(userData, res) {
-        if (!userData.loggedIn) {
-            sails.log.error('User is not logged in', userData);
-            return res.forbidden();
-        }
     },
 };
 
--- a/api/controllers/EAppEligibilityQuestionsController.js
+++ b/api/controllers/EAppEligibilityQuestionsController.js
@@ -1,0 +1,120 @@
+const EAppEligibilityQuestionsController = {
+    renderQuestionOne(req, res) {
+        return EAppEligibilityQuestionsController._pageRendering(
+            req,
+            res,
+            'eApostilles/eligibilityQuestionOne.ejs'
+        );
+    },
+
+    renderQuestionTwo(req, res) {
+        return EAppEligibilityQuestionsController._pageRendering(
+            req,
+            res,
+            'eApostilles/eligibilityQuestionTwo.ejs'
+        );
+    },
+
+    renderQuestionThree(req, res) {
+        return EAppEligibilityQuestionsController._pageRendering(
+            req,
+            res,
+            'eApostilles/eligibilityQuestionThree.ejs'
+        );
+    },
+
+    handleQuestionOneAnswer(req, res) {
+        const params = {
+            radioInputName: 'eapostille-acceptable',
+            errorPagePath: 'eApostilles/eligibilityQuestionOne.ejs',
+            redirectOptions: {
+                yes: '/eligibility-question-two',
+                no: '/use-standard-service',
+            },
+        };
+        return EAppEligibilityQuestionsController._handleYesNoAnswers(
+            req,
+            res,
+            params
+        );
+    },
+
+    handleQuestionTwoAnswer(req, res) {
+        const params = {
+            radioInputName: 'documents-eligible',
+            errorPagePath: 'eApostilles/eligibilityQuestionTwo.ejs',
+            redirectOptions: {
+                yes: '/eligibility-question-three',
+                no: '/use-standard-service',
+            },
+        };
+        return EAppEligibilityQuestionsController._handleYesNoAnswers(
+            req,
+            res,
+            params
+        );
+    },
+
+    handleQuestionThreeAnswer(req, res) {
+        const params = {
+            radioInputName: 'notarised-and-signed"',
+            errorPagePath: 'eApostilles/eligibilityQuestionThree.ejs',
+            redirectOptions: {
+                yes: '/eapp-start-page',
+                no: '/use-notarised-pdf',
+            },
+        };
+        return EAppEligibilityQuestionsController._handleYesNoAnswers(
+            req,
+            res,
+            params
+        );
+    },
+
+    _handleYesNoAnswers(req, res, params) {
+        const {radioInputName, errorPagePath, redirectOptions} = params;
+        const radioValueSelected = req.body[radioInputName];
+        const userData = EAppEligibilityQuestionsController._fetchUserData(
+            req,
+            res
+        );
+
+        if (!radioValueSelected) {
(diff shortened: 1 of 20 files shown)


assets/styles/informed.scss
```diff
@@ -1860,7 +1860,7 @@ margin-bottom: 10px
 
 .govuk-button--start {
   font-weight: 700;
-  font-size: 18px!important;
+  font-size: 24px!important;
   font-size: 1.125rem;
```
jenbutongit: should it be 24px or 1.125rem? There shouldn't be a need to use `!important`. Let's delete one of them and make sure the other is correct

It might also be worth using a rem size like govuk have. It looks like your root size is 16px, so `24/16` should give you the rem size

assets/styles/informed.scss
```diff
@@ -2069,3 +2069,7 @@ caption {
 .govuk-warning-text__text {
   font-size: 19px !important;
 }
+
+html {
+  font-size: 16px;
+}
```
jenbutongit: there's a lot of font size declaration in this. are we importing font sizes/globals from govuk-frontend? if so, do we need to keep these?

config/routes.js
```diff
@@ -79,8 +76,28 @@ module.exports.routes = {
...
+    '/handle-question-one-answer':
+        'EAppEligibilityQuestionsController.handleQuestionOneAnswer',
+    '/eligibility-question-two':
+        'EAppEligibilityQuestionsController.renderQuestionTwo',
+    '/handle-question-two-answer':
+        'EAppEligibilityQuestionsController.handleQuestionTwoAnswer',
+    '/eligibility-question-three':
+        'EAppEligibilityQuestionsController.renderQuestionThree',
+    '/handle-question-three-answer':
+        'EAppEligibilityQuestionsController.handleQuestionThreeAnswer',
```
jenbutongit: I don't know enough about sails.. but these look like routes. This might be easier over another call for some of the points below. 

-- 
URLs

govuk/gds have standards for URLs you might now have seen - https://www.gov.uk/guidance/content-design/url-standards-for-gov-uk. There's a bunch of info there as to why they suggest these.

A couple of the points to note 

> - URLs always need to be clear, unambiguous, easy to read, easy to type and easy to share 
> - URLs should align with the title of the page

I'd say it's "usually" up to a content designer to go in and clean up things like titles and URLs, but they're popular people and can be difficult to get a hold of... good to keep these things in mind in case that happens. If in doubt, kebab-case the question 😄 

-- 
/handle-x routes

It's not necessary to have separate routes here as far as I can see. One route can use multiple HTTP verbs. What would usually happen is 
1. `GET /question-1` - renders the page
2.  a `<form>`'s method can be set (i.e. how it sends the form to the server) to post. `<form method="post">`
3. user fills out the form and clicks continue 
4. the handler `POST /question-1` handles where to send the user next. You will have all the data the user sent in req.body 

judging by the route names, it looks like you've been inspired by react (i.e. `handleButtonClick`).. unfortunately this one doesn't translate so well to backend!

api/controllers/EAppEligibilityQuestionsController.js
```diff
@@ -0,0 +1,120 @@
...
+        );
+    },
+
+    renderQuestionThree(req, res) {
+        return EAppEligibilityQuestionsController._pageRendering(
+            req,
+            res,
+            'eApostilles/eligibilityQuestionThree.ejs'
+        );
+    },
```
jenbutongit: These routes look essentially the same. There are a few things most/all web frameworks will do to make this simpler/DRYer. (Suggest you read the comments on config/routes.js first since this will make more sense) Dynamic parameters will be useful here. 
https://sailsjs.com/documentation/concepts/routes/custom-routes#?wildcards-and-dynamic-parameters

using this (combined with comments in routes.js), I might have a route `/eligibility/:question`. So, when a user navigates to `service.gov.uk/eligibility/is-e-Apostille-accepted-in-destination`, I can do `req.param('question')`, and it will give me `is-e-Apostille-accepted-in-destination`. I can then combine with a condition object

```suggestion
    eligibilityQuestion(req, res) {
        const pages = {
          // "question-1" : "relevant-template.ejs" 
        }
        const question = req.param('question')
        return EAppEligibilityQuestionsController._pageRendering(
             req,
             res,
             pages[question]
         );
    },
```

it may not even be necessary to have this separate _pageRendering method if you're able to get the 3 methods down to one

api/controllers/EAppEligibilityQuestionsController.js
```diff
@@ -0,0 +1,120 @@
...
+        if (!radioValueSelected) {
+            sails.log.error('No option selected');
+            return res.view(errorPagePath, {
+                user_data: userData,
+                page_error: true,
+            });
+        }
+
+        return res.redirect(redirectOptions[radioValueSelected]);
+    },
```
jenbutongit: If you follow the comments on +2- +12 on this file, you could trim these down to one route (see route.js comment as well)

views/partials/header.ejs
```diff
@@ -116,14 +116,14 @@
         <h1 class='heading-large'>
             Your application will time out soon
         </h1>
-        <p>We will reset your application if you do not respond in</p>
-        <p style="margin: 0; padding: 0 10px; float: left;">
+        <p class="govuk-body">We will reset your application if you do not respond in</p>
+        <p class="govuk-body" style="margin: 0; padding: 0 10px; float: left;">
```
jenbutongit: doesn't look like your code, but it's worth cleaning up the inline styles for peace of mind

views/eApostilles/startPage.ejs
```diff
@@ -1,5 +1,5 @@
 <div class="inner_header no-user-signed-in">
-    <a href="/select-service" class="back-to-previous govuk-link">Back</a>
+    <a href="/eligibility-question-three" class="back-to-previous govuk-link">Back</a>
```
jenbutongit: would a user be able to go straight to this page? could you always guarantee the only way to see this content is via /eligibility-question-three?

jenbutongit (changes requested):

assets/styles/informed.scss
RichardBray: Good shout. I've noticed the `govuk-button--start` class is there twice, one added manually and one imported from govuk-frontend, so I've gotten rid of the manual one. Hope it doesn't break anything somewhere else 🤞

assets/styles/informed.scss
RichardBray: So the css in this project is a bit all over the place. It looks like there do import the base gov css like so:
```
@import "../../node_modules/govuk-elements-sass/public/sass/elements/_govuk-template-base"; 
```
which seems to have html font-size as this:
```
html {
  font-size: 62.5%; // 1
...
```

Which doesn't look right to me. I'll see if I can update the base.

api/controllers/EAppEligibilityQuestionsController.js
RichardBray: This is a really good point. I've never thought about passing dynamic values through the url for this but it does make a lot of sense. Hopefully shouldn't be too difficult to implement :)

config/routes.js
RichardBray: I actually didn't know you could use REST verbs in sailsjs because I hadn't seen it in this project but looks like you can. Can't imagine why they weren't used before 🤔

api/controllers/EAppEligibilityQuestionsController.js
```diff
@@ -0,0 +1,100 @@
...
+        const eligibilityViews = {
+            'apostille-accepted-in-desitnation':
+                'eApostilles/eligibilityQuestionOne.ejs',
+            'documents-eligible-for-service':
+                'eApostilles/eligibilityQuestionTwo.ejs',
+            'pdfs-digitally-signed':
+                'eApostilles/eligibilityQuestionThree.ejs',
+        };
+
+        const questionPage = eligibilityViews[req.param('question')];
```
jenbutongit: ~~if a user manually inputs into their url bar "....gov.uk/eligibility/something-that-doesnt-exist" - will it show a user a 404? I'm suspecting that it might end up giving the user a 500 (internal server error) which wouldn't be correct. Maybe default to the first page?~~

nevermind, this is defined in config/routes so sails should 404 it, feel free to dismiss

config/routes.js
```diff
@@ -79,8 +76,20 @@ module.exports.routes = {
         'CheckUploadedDocumentsController.addDocsToDBHandler',
     '/handle-service-choice': 'ApplicationTypeController.handleServiceChoice',
 
+    // eligibility questions
+    'GET /eligibility/:question':
+        'EAppEligibilityQuestionsController.renderEligibilityQuestion',
+    'POST /eligibility/:question':
+        'EAppEligibilityQuestionsController.handleEligibilityAnswers',
```
jenbutongit, line 83: 💪

views/partials/header.ejs
```diff
@@ -116,14 +116,14 @@
...
+        <p class="govuk-body">We will reset your application if you do not respond in</p>
+        <p class="govuk-body session-expiry-popup__clock">
             <img class='countdown-clock' src='/images/icon-countdown-clock.png' alt='Countdown clock'/>
-        <div style="margin-top: 1.3em;">
+        <div class="govuk-body popup__countdown">
             <span id="displayTime"></span>
         </div>
         </p>
-        <p>We do this to keep your information secure.</p>
+        <p class="govuk-body">We do this to keep your information secure.</p>
```
jenbutongit: sorry, I didn't spot this before, but we should use these (you'll have to convert the template to ejs). It has the right roles and aria attributes so that it can be read out nicely for someone using a screen reader

https://github.com/XGovFormBuilder/digital-form-builder/blob/9953ff825ed7e205c63374fa6906d89860517e0a/runner/public/static/modal-dialog.js

https://github.com/XGovFormBuilder/digital-form-builder/blob/9953ff825ed7e205c63374fa6906d89860517e0a/runner/src/server/views/partials/modal-dialog.html

jenbutongit (changes requested):

api/controllers/EAppEligibilityQuestionsController.js
RichardBray: Yeah looks like it's taken care of by the Sailsjs 404 page. Not the best 404 message but it's something 🤷 

Screenshot 2021-09-02 at 18 07 22

Actually, looks like that's the 403 page. This project is so random 🤦

RichardBray: Okay, should be fixed to go to 404 now.

Screenshot 2021-09-02 at 18 33 46

views/partials/header.ejs
RichardBray: Maybe we should have a chat about this change tomorrow. I've had a look through the markup and I have a few questions about how some of the attributes work like `data-minutes-idle-timeout` and `data-minutes-modal-visible`.

Also, it looks like the HTML dialog element doesn't have great browser support :(
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

Maybe there's a fallback you use in the forms project or something 🤷

jenbutongit: They will still render since they've got styles (although you might need to dig them out? I forget) and `role="dialog"` will cover any browsers that don't support it. 

As for the attributes, they are used by the js file https://github.com/XGovFormBuilder/digital-form-builder/blob/9953ff825ed7e205c63374fa6906d89860517e0a/runner/public/static/modal-dialog.js#L25
`idleMinutesBeforeTimeOut: $('#js-modal-dialog').data('minutes-idle-timeout')`

RichardBray: Yeah, I noticed the attributes in that file, haven't seen them used like that before, with data attributes but it makes sense.

jenbutongit: it's a jquery-ism _shudder_

jenbutongit (approved):

PR #277, opened 2021-09-09
Feature/EIA-105 - 'Available to download' countdown

RichardBray:
# Description

The purpose of this ticket is to visually display to the user how many days they have left to download their application based on the current date. The user has 21 days to download their application after it has been fully processed

---

Screenshot 2021-09-10 at 11 03 41

---

Due to the missing application download endpoint from Iizuka, the screenshot above displays the notification on an in-progress application, but that is for testing purposes only. The in-progress application page should not contain the notification code.

**https://jira.digital.csd.fcdo.gov.uk/browse/EIA-105**

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:


api/controllers/OpenEAppController.js
```diff
@@ -96,7 +100,31 @@ const OpenEAppController = {
     },
 
     _formatDate(date) {
-        return moment(date).format('DD MMMM YYYY');
+        return dayjs(date).format('DD MMMM YYYY');
+    },
+
+    _calculateDaysLeftToDownload(req) {
+        const urlObj = new URL(
+            `${req.protocol}://${req.headers.host}${req.originalUrl}`
```
jenbutongit: does this mean you'll accept HTTP requests too?

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +100,31 @@ const OpenEAppController = {
...
+        }
+        const completedDate = queryParams.get('completedDate');
+        const currentDate = dayjs(Date.now());
+        const differenceBetweenCurrentAndCompletedDate = currentDate.diff(
+            completedDate,
+            'day'
+        );
+        const daysLeftToDownloadDoc =
+            MAX_DAYS_TO_LEGALISE -
+            Number(differenceBetweenCurrentAndCompletedDate);
```
jenbutongit: why does `differenceBetweenCurrentAndCompletedDate` get cast to number via `Number`?  day.js documentation says this 

>By default, dayjs#diff will truncate the result to zero decimal places, returning an integer 

so it doesn't look like it needs to be wrapped in Number.

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +100,31 @@ const OpenEAppController = {
...
+        const numberIsNegative = Math.sign(daysLeftToDownloadDoc) === -1;
+        if (numberIsNegative) {
+            throw new Error('Application has expired');
+        }
```
jenbutongit: maybe this?
```suggestion
        if (daysLeftToDownloadDoc < 0) {
            throw new Error('Application has expired');
        }
```

I appreciate verbosity when naming things, it makes things easier to read - but this might be too verbose or even adding complexity where there doesn't need to be. `< 0` is simpler (and less lines ;))

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +100,31 @@ const OpenEAppController = {
...
+        return daysLeftToDownloadDoc;
```
jenbutongit: I'm just having a look at day.js documentation, and I think if we use some of their methods it will be a little simpler overall.

day.js has a few neat features that we can leverage so you don't have to write it all out, and is probably well tested (i.e. less for you to write ;)) https://day.js.org/docs/en/durations/diffing or https://day.js.org/docs/en/durations/subtract

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +100,31 @@ const OpenEAppController = {
...
     _formatDate(date) {
-        return moment(date).format('DD MMMM YYYY');
+        return dayjs(date).format('DD MMMM YYYY');
+    },
+
+    _calculateDaysLeftToDownload(req) {
+        const urlObj = new URL(
+            `${req.protocol}://${req.headers.host}${req.originalUrl}`
+        );
+        const queryParams = new URLSearchParams(urlObj.search);
```
jenbutongit: what stops a user from changing the queryParams themselves so they're able to get to their documents? or is that not a risk?

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +100,31 @@ const OpenEAppController = {
...
+            throw new Error('No date value found in url');
+        }
+        const completedDate = queryParams.get('completedDate');
+        const currentDate = dayjs(Date.now());
+        const differenceBetweenCurrentAndCompletedDate = currentDate.diff(
+            completedDate,
+            'day'
+        );
+        const daysLeftToDownloadDoc =
+            MAX_DAYS_TO_LEGALISE -
```
jenbutongit: should it be `MAX_DAYS_TO_DOWNLOAD`? or `daysLeftToLegalise`? it's not clear that these two variables should be or are related since their names are different

tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -148,4 +162,56 @@ describe('OpenEAppController', () => {
...
+            sandbox.stub(Date, 'now').callsFake(() => TWO_DAYS_AFTER_DEADLINE);
+            await OpenEAppController.renderPage(reqStub, resStub);
+
+            // then
+            const sailsErrorLogObj = sails.log.error.getCall(0).args[0];
+            expect(sailsErrorLogObj.message === 'Application has expired').to.be
+                .true;
+            expect(resStub.serverError.called).to.be.true;
+        });
+        it('shows correct number of days for 11 day old application', async () => {
```
jenbutongit, line 172: what happens when the deadline is in 1 hour? (0 days, 1hr?) - would you just show 0 days to download?

tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -148,4 +162,56 @@ describe('OpenEAppController', () => {
...
+            // when
+            sandbox.stub(Date, 'now').callsFake(() => ELEVEN_DAYS_BEFORE_DEADLINE);
+            await OpenEAppController.renderPage(reqStub, resStub);
+
+            // then
+            expectedPageData.daysLeftToDownload = 10;
```
jenbutongit: should this be an `expect`? it is just being assigned as 10 here. is that right?

tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -148,4 +162,56 @@ describe('OpenEAppController', () => {
             );
         });
     });
+
+    describe('date countdown', () => {
+        beforeEach(async () => {
```
jenbutongit, line 163: `OpenEAppController. _calculateDaysLeftToDownload` is quite an important method here. I think there should be some unit tests to make sure it does return (or throw) what you're expecting.

jenbutongit (changes requested):

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Yep, currently, it's just 0 days but there's talk of changing the copy of some sort

api/controllers/OpenEAppController.js
RichardBray: yep, currently testing locally only has HTTP so I'll have to test on the AWS instance to see if it'll do HTTPS as well.

api/controllers/OpenEAppController.js
RichardBray: Good point. Nothing stops the user from changing the query params, the query param will only be used for the notification, changing it won't give users access to their documents if the download has expired.

But I've just had a thought which means I might be able to do it without a query param. I'm going to try it out :)

api/controllers/OpenEAppController.js
RichardBray: I admit it's a bit confusing. I'll fix that 👍

api/controllers/OpenEAppController.js
RichardBray: Good spot, this is some legacy code. I wasn't using the dayjs diff method before, the method I was using before returned a string. But I can remove that now.

api/controllers/OpenEAppController.js
RichardBray: It's so simple, I don't know why I didn't think of it before 😭

jenbutongit: You did submit the PR on a Friday 😄

tests/specs/controllers/OpenEAppController.test.js
jenbutongit: do you have i18n or a way of handling "1 day" vs "0 days" vs "20 days" etc?

api/controllers/OpenEAppController.js
RichardBray: ha, that's true

api/controllers/OpenEAppController.js
RichardBray: Good shout, actually if we start doing days and hours for some reason I think using their methods makes sense.

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Hmmm, I don't have i18n but I don't follow what you mean by handling different days. Are you referring to different languages? Or something else?

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Not quite, so I'm passing in a date 11 days before the deadline and expecting the page data to return 10 days. If the 10 in this line was changed the tests would fail, well should fail.

Re-reading that I'm not sure if it makes sense 🤔

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Yeah, this is a good point. Some of my other tests sort of test this function already since this needs to return the right value for the others to pass but I'm happy to add specific tests for this function as well?

tests/specs/controllers/OpenEAppController.test.js
jenbutongit: i18n is used for languages but most of them have a way of handling variations of words, like plural or singular. So, when you pass "0" does the user see 0 days remaining? what about if you pass 1, does the user see "1 days remaining"?  or do they see 1 day remaining?

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +104,29 @@ const OpenEAppController = {
...
+            applicationTableData.createdAt
+        );
+        const differenceAsDayjsObj = dayjs.duration(
+            differenceBetweenCurrentAndCompletedDate
+        );
+        const daysLeftToDownloadDoc = dayjs
+            .duration(MAX_DAYS_TO_DOWNLOAD, 'd')
+            .subtract(differenceAsDayjsObj)
+            .days();
+        console.log(daysLeftToDownloadDoc, 'daysLeftToDownloadDoc');
```
jenbutongit: console.log

tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -148,4 +156,91 @@ describe('OpenEAppController', () => {
...
+            const expectedValues = [9, 14, 19, 0];
+            const returnedValues = currentDates.map((currentDate) => {
+                sandbox.stub(Date, 'now').callsFake(() => currentDate);
+                const val = OpenEAppController._calculateDaysLeftToDownload({
+                    createdAt: resolvedAppData.createdAt,
+                });
+                Date.now.restore();
+                return val;
+            });
+            console.log(returnedValues, 'returnedValues');
```
jenbutongit: console.log

tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -148,4 +156,91 @@ describe('OpenEAppController', () => {
...
+            expect(resStub.serverError.called).to.be.true;
+        });
+    });
+
+    describe('_calculateDaysLeftToDownload', () => {
+        it('returns expected values', () => {
+            // when
+            const TWELVE_DAYS_AFTER_COMPLETION = 1630281600000;
+            const SEVEN_DAYS_AFTER_COMPLETION = 1629849600000;
+            const TWENTY_ONE_DAYS_AFTER_COMPLETION = 1631142000000;
```
jenbutongit, line 208: what happens when you pass 22 days after completion?

jenbutongit (changes requested):

tests/specs/controllers/OpenEAppController.test.js
RichardBray: So it should respond with an error saying the application is expired. The test for that is up here on line 170:

https://github.com/UKForeignOffice/loi-application-service/pull/277/files/3d98dd43b1bfc655d0a608f6ffee8caf3d3ad0c3#diff-fef66b6a716dd8342942b557f6df57eaa714e1484126933f96ebd974aa48a980R170-R182

But I'm happy to move it to this section, or move this section above if you think it makes sense.

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Yeah good shout, I think they do. But I don't know if it's worth downloading a whole library just for that one change 🤔

jenbutongit: It might not be, but maybe an if (or ternary) statement in the ejs file will suffice?

RichardBray:
>i18n is used for languages but most of them have a way of handling variations of words, like plural or singular. So, when you pass "0" does the user see 0 days remaining? what about if you pass 1, does the user see "1 days remaining"? or do they see 1 day remaining?

For some reason, I wasn't able to reply to this comment above so I'm replying here.

Ah, that's cool, never used i18n for plurals before, although I kind of feel bad installing a whole new library just for that.

Also now that I'm thinking about it, this ticket wasn't very well sliced, but that's more down to Iizuka than anyone at CYB. Basically, the date notification countdown thing can only be shown on a completed application. The screenshot above is from an in-progress app that I used to test the countdown.

Iizuka hasn't yet created the endpoint to download documents for a completed app, so although I've created the page and have some dummy data using the documentation they provided, the work won't be 'done' until they've done their bit. So this ticket is more logic for now than UI. The UI will be done and ready for review once the endpoint from Iizuka is done.

Happy to have a call to explain this further in a call if you want. But the plural stuff won't be done in this ticket.

jenbutongit:
on installing libraries, it's not a huge deal. Eventually you would have to install one to support welsh. 

Roger doger on iizuka - ignore my requests then

tests/specs/controllers/OpenEAppController.test.js
jenbutongit: `calculateDaysLeftToDownload` is more predictable to test (i.e. "pure".. or at least slightly more "pure"). It would be faster (in terms of execution, and writing the tests) to just test that if you're strapped for time. And the render method relies on it. By testing the render function you're having to spy the sails method.

```.js
expect(sailsErrorLogObj.message === 'No date value found').to.be
                 .true;
``` 
is the same message that `calculateDaysLeftToDownload` throws. 

usually if I am testing a render method, I'd be checking if what it renders on a page (or the response body) contains what I'm expecting it to

jenbutongit (changes requested):

RichardBray:
I wonder why I can't reply to your comments anymore 🤔 

> usually if I am testing a render method, I'd be checking if what it renders on a page (or the response body) contains what I'm expecting it to

Cool, I'm pretty new to writing backend tests so thanks for this tip. I'm assuming you mean you test the response methods and arguments? When I hear "render method" it makes me think of React :)

Then again I think everyone does things differently. When I was writing tests for ETD Ali got me to tests the logs were printing out the correct text. I think I'll write tests for both things to be on the safe side 😁

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +105,26 @@ const OpenEAppController = {
...
     _formatDate(date) {
-        return moment(date).format('DD MMMM YYYY');
+        return dayjs(date).format('DD MMMM YYYY');
+    },
+
+    _calculateDaysLeftToDownload(applicationTableData) {
+        if (!applicationTableData.createdAt) {
+            throw new Error('No date value found');
+        }
+        const todaysDate = dayjs(Date.now());
```
jenbutongit, line 115: `dayjs()` has the same behaviour as `Date()` (creates a Date with the datetimestamp of "now" --- but is different to Date.now() which returns it in unix epoch in ms) - so I think it's fine to just have it as `dayjs()`

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +105,26 @@ const OpenEAppController = {
...
+        const differenceBetweenCurrentAndCompletedDate = todaysDate.diff(
+            applicationTableData.createdAt
+        );
```
jenbutongit, line 118: maybe `timeSinceCompletedDate`?

api/controllers/OpenEAppController.js
```diff
@@ -96,7 +105,26 @@ const OpenEAppController = {
...
+        const maxDaysToDownload = dayjs.duration({
+            days: MAX_DAYS_TO_DOWNLOAD,
+        });
+        const differenceAsDayjsObj = dayjs.duration(
```
jenbutongit, line 122: you are only using dayjs objects here so I don't think it makes sense to use asDayjsObj

jenbutongit (approved):
lgtm, minor comments if you've got time to action them, mostly on readability

api/controllers/OpenEAppController.js
RichardBray: Yeah I did originally have it as just `days()` but I needed a way to stub the current date so I could test the countdown is returning the right value, so I'm stubbing the Date.now method instead. Don't know if this is the best solution but I'm open to options.

api/controllers/OpenEAppController.js
RichardBray: much better, I struggle naming variable sometimes :)

api/controllers/OpenEAppController.js
RichardBray: makes sense

PR #278, opened 2021-09-10
Feature/EIA-94 Internal reference (aka Your reference)

RichardBray:
# Description

 **1 - VIDEO RUN-THROUGH 🎨**

https://user-images.githubusercontent.com/1377253/133070303-6122e1cd-296c-4d25-9fe1-57df44a81708.mp4

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-94

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:


api/models/ExportedEAppData.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+module.exports = {
+    attributes: {
+        application_id: {
+            type: Sequelize.INTEGER(),
+        },
+        applicationType: {
+            type: Sequelize.STRING(),
+        },
```
jenbutongit, line 13: is it too late to change `applicationType`? it's inconsistent with the other attributes here

api/controllers/EAppReferenceController.js
```diff
@@ -0,0 +1,29 @@
+const sails = require('sails');
+
+const EAppReferenceController = {
+    renderPage(req, res) {
+        const userData = HelperService.getUserData(req, res);
+        if (!userData.loggedIn) {
+            sails.log.error('User not logged in');
+            return res.serverError();
```
jenbutongit: what happens when a user isn't logged in? is a generic message displayed? or are they able to find out how to log in from there?

tests/specs/controllers/EAppReferenceController.test.js
```diff
@@ -0,0 +1,92 @@
...
+            // when
+            sandbox.stub(HelperService, 'getUserData').callsFake(() => ({
+                loggedIn: false,
+            }));
+            EAppReferenceController.renderPage(reqStub, resStub);
+
+            // then
+            expect(
+                sails.log.error.getCall(0).args[0] ===
+                    'User not logged in'
```
jenbutongit: Could you think of a way to test without relying on spying on the log? When implementation details are tested, the tests end up being flakier (likely to break down the line), and that will affect time it takes for a feature to be "done" (since you have to go back and fix the tests). 

it looks like spying on sails.log is a workaround for testing this "error" block in `EAppReferenceController.renderPage`.  Ideally there shouldn't be a need to test a 3rd party library (i.e. sails.log) because it's already well tested. You may need to spy and test on sails.log once in your application to make sure it can be called.

from Kent C Dodds ((react) testing-library author) 
>Implementation details are things which users of your code will not typically use, see, or even know about.

Here is his blog post about it (however this one is more about front end testing https://kentcdodds.com/blog/testing-implementation-details) 

You sort of have to imagine you only know the parameters of the method (req, res), does the method return what you want it to? and if renderPage requires too much stubbing/spying, how could you test smaller chunks of the code?

If a method doesn't return anything, yes you may have to resort to spying on a log 

p.s. spying is also useful for testing on old systems you don't have time to refactor, but imo should only be a last resort (or the alternative is too painful)

tests/specs/controllers/EAppReferenceController.test.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    describe('addReferenceToSession', () => {
+        it('updates session with new userRef and redirects', () => {
+            // when
+            reqStub.body['user-reference'] = 'TestRef';
+            EAppReferenceController.addReferenceToSession(reqStub, resStub);
+
+            // then
+            expect(reqStub.session.eApp.userRef === 'TestRef').to.be.true;
```
jenbutongit: should this be `expect(reqStub.session.eApp.userRef).to.be('TestRef')`? any reason why `===`?

tests/specs/controllers/EAppReferenceController.test.js
```diff
@@ -0,0 +1,92 @@
...
+                .be.true;
+        });
+
+        it('logs error if no userRef found', () => {
+            // when
+            reqStub.body['user-reference'] = null;
+            EAppReferenceController.addReferenceToSession(reqStub, resStub);
+
+            // then
+            expect(sails.log.error.calledWith('No user reference found')).to.be.true;
```
jenbutongit: so, following from above.. this test is only useful if you know this error message will be used to debug someone's session in production. Nobody can see the logs apart from kainos. 

So, if a user is missing a user-reference, how do you know something hasn't been added to the session? Should null or undefined be added to the session? What if the user-reference was `1`?

It's more helpful to test the "output" or the thing that changes

jenbutongit (changes requested):
A few comments on the testing - let me know if any Qs

tests/specs/controllers/EAppReferenceController.test.js
RichardBray: No reason, I was just experimenting with using fewer matchers, so just true and false. But it's not that big of a deal to add equals as well.

api/models/ExportedEAppData.js
RichardBray: As in using camel case instead of snake case? Erm, no it's not too late to change, would be a bit of a faff but should be doable. Although my inspiration was from this file:

https://github.com/UKForeignOffice/loi-application-service/blob/develop/api/models/ExportedApplicationData.js#L12

They are pretty much doing the same thing except one is for electronic apps and one is for paper apps.

api/controllers/EAppReferenceController.js
RichardBray: Good point. It's quite interesting, the sign-in is handled by a completely different repo so it would have been nice to redirect to the login page with a message saying 'you don't have access to this page, but that ability doesn't exist :(

The res.forbidden() might be the next best thing, but I don't think that's configured to render a forbidden page 🤔

tests/specs/controllers/EAppReferenceController.test.js
RichardBray: That makes sense, I think for me it's always easier to figure out how to write FE tests because they're more visual but BE tests I think I need to get the hang of.

Also, I know you mentioned this in the previous PR, hopefully, I'll have this down in the next PR I create :)

api/controllers/EAppReferenceController.js
RichardBray: Actually, looks like forbidden redirects to the 403 page:

https://github.com/UKForeignOffice/loi-application-service/blob/develop/api/responses/forbidden.js#L51

which... has the same text as the 500 and 404 pages 🤦 

Ah well 🤷

tests/specs/controllers/EAppReferenceController.test.js
```diff
@@ -0,0 +1,86 @@
+const { expect } = require('chai');
+const sinon = require('sinon');
+
+const EAppReferenceController = require('../../../api/controllers/EAppReferenceController');
+
+describe.only('EAppReferenceController', () => {
```
jenbutongit: remove .only

api/models/ExportedEAppData.js
jenbutongit: oh I just noticed there's already a few using camelCase.. maybe add this as a tech debt/refactor ticket for a future date. Ideally they're all the same case so we always treat `attributes` the same

jenbutongit (changes requested):
lgtm, just need get rid of the .only on the test case 🚀

RichardBray:
> oh I just noticed there's already a few using camelCase.. maybe add this as a tech debt/refactor ticket for a future date. Ideally they're all the same case so we always treat attributes the same

Yeah that makes sense. I can make a ticket to sort that out.

jenbutongit (approved):

PR #280, opened 2021-09-15
bug/EIA-174 Fix issue with upload limits on document upload

RichardBray:
# Description

https://user-images.githubusercontent.com/1377253/134202781-d9a2f3e6-a0d5-4dc2-95cd-dff5295c81e1.mp4

This is a bug ticket that fixes an issue where upload limits are being calculated as an aggregate instead of per file.

Below is a screenshot showing what happens when two large pdfs (190MB) are uploaded with the new change, they are deleted and this error message is presented.

---

Screenshot 2021-09-21 at 16 44 10

---

Here is the link from the video with the multer issue --> https://github.com/expressjs/multer/issues/344

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-174

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:

### DIFF
--- a/tests/specs/helpers/uploadedFileErrorChecks.test.js
+++ b/tests/specs/helpers/uploadedFileErrorChecks.test.js
@@ -1,74 +1,125 @@
-const { expect } = require("chai");
-const sinon = require("sinon");
+const { expect } = require('chai');
+const sinon = require('sinon');
+const fs = require('fs');
 const {
-  checkTypeSizeAndDuplication,
-} = require("../../../api/helpers/uploadedFileErrorChecks");
+    checkTypeSizeAndDuplication,
+    displayErrorAndRemoveLargeFiles,
+} = require('../../../api/helpers/uploadedFileErrorChecks');
 
+const sandbox = sinon.sandbox.create();
 
-describe("checkTypeSizeAndDuplication", () => {
-  const sessionStub = (uploadedFiles) => ({
-    eApp: {
-      uploadedFileData: uploadedFiles,
-      uploadMessages: {
-        errors: [],
-      },
-    },
-  });
-  const requestStub = (uploadedFiles = []) => ({
-      session: sessionStub(uploadedFiles),
-      headers: {
-          'content-length': 136542,
-      },
-  });
+describe('checkTypeSizeAndDuplication', () => {
+    afterEach(() => {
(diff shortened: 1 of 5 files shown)


PaulioRandall:
Regarding the issue of file size detection not possible until after the file has been fully uploaded, from UX perspective it would be good to avoid uploading something we know will fail if we can, especially if the files take some time to upload. Then again, if most files are only a few megabytes then it probably won't provide much value. If you were to tackle that issue, I'd probably create another ticket for it.

tests/specs/helpers/uploadedFileErrorChecks.test.js
```diff
@@ -1,74 +1,124 @@
...
+            },
+        };
+        displayErrorAndRemoveLargeFiles(reqStub);
+
+        // then
+        expect(reqStub.session.eApp.uploadedFileData.length).to.equal(0);
+        expect(reqStub.session.eApp.uploadMessages.errors.length).to.equal(2);
+        expect(unlinkFile.callCount).to.equal(2);
+    });
 });
```
PaulioRandall: Might be worth another test to ensure only the oversized files are removed from the set of upload files, or in other terms, if only one file is too large then the other files are uploaded successfully. Apologise if I've mis-understood the situation.

RichardBray:
> Regarding the issue of file size detection not possible until after the file has been fully uploaded, from UX perspective it would be good to avoid uploading something we know will fail if we can, especially if the files take some time to upload. Then again, if most files are only a few megabytes then it probably won't provide much value. If you were to tackle that issue, I'd probably create another ticket for it.

Yep, the hope is that only a few megabytes would be loaded in for the private beta, we do have a 200Mb limit per file but I highly doubt someone would upload a document that large.

Actually, there is a ticket to add a drag and drop feature for those who have JS enabled. The size checking on the front end could definitely be done in that ticket 🤔

PaulioRandall:
I can't run the tests myself but the code for the new test looks good.

RichardBray:
> I can't run the tests myself but the code for the new test looks good.

Fair enough, thanks for the approval 👍

PR #285, opened 2021-09-29
Feature/EIA-104 Download documents individually when viewing completed application

RichardBray:
# Description

 **Video to show feature working**

https://user-images.githubusercontent.com/1377253/135310767-bf884301-aa82-4a20-b823-0d5dd3af44cb.mp4

This PR adds the ability to download legalised documents from Casebook.
Casebook returns the document as a stream which is sent directly to the client once the download button is clicked.

This PR also adds some download-specific copy to the view which is based on the application and document status sent from Casebook.

An application can have one of three statuses: **In Progress, Quality Assurance, and Done**.
A document can only be downloaded if an application is **Done**.

A document on the other hand has four statuses: **Submitted, Checked, Rejected, and Fraudulent**.

**Then Rejected and Fraudulent statuses will be dealt with in a follow-up PR.**

A document can only be downloaded if it is **Checked**.

Downloads only work when Casebook returns the **apostilleReference** value for a document, which only happens when an application is **Done** and a document is **Checked**. Hence the nested if statements in the .ejs files.

**Jira ticket** https://jira.digital.csd.fcdo.gov.uk/browse/EIA-104

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:


config/routes.js
```diff
@@ -63,6 +63,10 @@ module.exports.routes = {
     '/check-documents-certified/confirm':
         'DocumentsCheckerController.docsCertifiedCheckConfirmDeny',
 
+    ////////////////////////////
+    // ---- e-Apostille ---- //
+    //////////////////////////
```
jenbutongit, line 68: damn, retro

api/controllers/FileDownloadController.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        const {
+            hmacKey,
+            customURLs,
+            casebookCertificate: cert,
+            casebookKey: key,
+        } = req._sails.config;
+        const queryParamsObj = {
+            timestamp: new Date(Date.now()).getTime().toString(),
```
jenbutongit: you don't have to pass in `Date.now()` to `new Date(..)`.  `new Date()` (with no params) will return you a Date object using the date time that it was called. 

`Date.now()` will return you epoch time, so it looks like you just need `Date.now().toString()` (or use a template string) 

repl below - they return the same value 
````
> console.log(new Date(Date.now()).getTime().toString(), '\n', new Date().getTime().toString(), '\n', Date.now().toString(), '\n', `${Date.now()}`)
1632936724048 
 1632936724048 
 1632936724048 
 1632936724048
````

File: api/controllers/FileDownloadController.js [hunk 3]
```diff
@@ -0,0 +1,66 @@
...
+                hash,
+            },
+            qs: queryParamsObj,
+        };
+    },
+
+    _streamFileToClient(options, res) {
+        sails.log.info('Downloading file from Casebook');
+        request
+            .get(options)
```
jenbutongit, line 55: I think you mentioned on engineering it's a deprecated library. Let's avoid if possible. I think it's ok to leave it as it is, unless npm reports a vulnerability related to this.

You may be able to use the native nodejs `http` client instead? 

https://nodejs.org/api/http.html#http_http_request_url_options_callback

tests/specs/controllers/FileDownloadController.test.js
```diff
@@ -0,0 +1,87 @@
...
+        };
+        const actualResult = FileDownloadController._prepareAPIOptions(reqStub);
+
+        // then
+        expect(actualResult).to.deep.equal(expectedResult);
+    });
+
+    it('streams file from the Casebook API to the client', () => {
+        // when
+        sandbox.stub(request, 'get').callsFake(() => ({
```
jenbutongit, line 71: don't worry about this for now but just as an FYI there are libraries out there that will help you mock a server, so instead of stubbing `request.get`, you are setting up your own server within the test for your `request` to interact with. That way you keep as much of your app unmocked. Generally looks nicer than stubbing on:pipe:on too 😄  https://github.com/nock/nock

tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -148,7 +149,7 @@ describe('OpenEAppController', () => {
             );
         });
 
-        it('should render openEApp.ejs page with correct data', () => {
+        it('should render openEA§pp.ejs page with correct data', () => {
```
jenbutongit: did you mean to do this?

tests/specs/controllers/FileDownloadController.test.js
```diff
@@ -0,0 +1,87 @@
...
+            serverError: sandbox.stub(),
+        };
+        sandbox.stub(Date, 'now').callsFake(() => 1483228800000);
+    });
+
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    it('throws an error if the apostilleRef param is undefined', () => {
```
jenbutongit, line 38: nice! 👍

jenbutongit (changes requested):
LGTM Rich - nice tests! Just a couple comments that don't need actioning, mainly an fyi for next time (or another project). Take a look at the timestamp comment though 🙂

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Ah, that's my bad. I'll update this 👍

tests/specs/controllers/FileDownloadController.test.js
RichardBray: Ah, that's cool, I didn't know about it. I'll bookmark this. Do you have any examples of this being used in other CYB projects you've worked on? If you've used this library before?

config/routes.js
RichardBray: 🤔

api/controllers/FileDownloadController.js
RichardBray: Good point. I think you've also mentioned this before :)

It's a big change but it makes sense. I'll create a ticket for it.

jenbutongit: you can always do it bit by bit, or for new features use http.request. Doesn't have to be all in one go, especially if there aren't any critical security issues with `request`

jenbutongit (approved):

PR #286, opened 2021-09-30
Merge Demo to UAT

RichardBray:


api/controllers/ApplicationTypeController.js
```diff
@@ -42,72 +51,102 @@ module.exports = {
...
+        const userLoggedIn = HelperService.LoggedInStatus(req);
+
+        if(!userLoggedIn) {
+            sails.log.error('User is not logged in');
+            return res.view('404');
+        }
+
+        const servicePages = {
+            eApostille: '/new-application?app_type_group=4',
+            standard: '/new-application?app_type_group=1',
```
simonmoore100, line 136: There is the drop off service as well which would be app_type_group=3

views/eApostilles/useStandardService.ejs
```diff
@@ -0,0 +1,28 @@
+<% pageTitle="You cannot use this service" %>
```
simonmoore100, line 1: When you hit this page, you loose the 'logged in' breadcrumb links for Applications, Account, Addresses and Sign out

views/eApostilles/applicationType.ejs
```diff
@@ -0,0 +1,150 @@
...
+                <ul class="govuk-list govuk-error-summary__list">
+                    <li>
+                        <a href="#no-service-chosen-error">You must select a service type.</a>
+                    </li>
+                </ul>
+            </div>
+        </div>
+    <%}%>
+
+    <h1 class="heading-large">Which service would you like?</h1>
```
simonmoore100, line 26: If you have the electronic flag enabled in the database, do you only get the option of e-App or standard journeys now? What about Premium or Drop-off?

views/eApostilles/applicationType.ejs
```diff
@@ -0,0 +1,150 @@
+<% pageTitle="Choose a service - Get a document legalised – GOV.UK" %>
```
simonmoore100, line 1: The 'Applications' breadcrumb link on this page doesn't work

edit: It does work but if casebook is down the page shows 'Page not found' error. The dashboard needs to render regardless of casebook connectivity

views/eApostilles/startPage.ejs
```diff
@@ -0,0 +1,31 @@
+<div class="inner_header no-user-signed-in">
```
simonmoore100, line 1: When you hit this page, you loose the 'logged in' breadcrumb links for Applications, Account, Addresses and Sign out

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -0,0 +1,270 @@
...
+    }
+}
+
+function initialiseClamScan(req) {
+    const { clamav_host: clamavHost, clamav_port: clamavPort } =
+        req._sails.config.eAppS3Vals;
+
+    const clamAvOptions = {
+        clamdscan: {
+            host: inDevEnvironment ? '127.0.0.1' : clamavHost,
```
simonmoore100: I think this part is unnecessary. If in dev mode, you aren't guaranteed to run on 127.0.0.1. I have my clamav running on a different docker network and need to comment this bit out to force it to use clamavHost.

Using the value from the env file should be fine in all cases.

views/eApostilles/uploadFiles.ejs
```diff
@@ -0,0 +1,164 @@
+<% const {uploadedFileData, uploadMessages} = req.session.eApp %>
+<% const {errors, infectedFiles, fileCountError} = uploadMessages %>
+
+<% pageTitle = (errors.length > 0 ? 'Error:' : '') + "Add your PDFs - Get a document legalised – GOV.UK" %>
+
+<section class="column-two-thirds">
+
+    <div class="inner_header <% if (!user_data.loggedIn){ %>no-user-signed-in<% } %>">
+        <a href="<%= backLink %>" class="back-to-previous govuk-link">Back</a>
```
simonmoore100: The breadcrumb links on this page need to be outside of this column-two-thirds div as they need to move over to the right a bit
Screenshot 2021-10-04 at 11 51 40
t

api/controllers/EAppSubmittedController.js
```diff
@@ -0,0 +1,180 @@
...
+
+        return {
+            application_id: sessionData.appId,
+            uploaded_url: fileUrl,
+            filename: uploadedFile.filename,
+        };
+    },
+
+    _generateS3PresignedUrl(uploadedfileName, s3Bucket) {
+        const EXPIRY_HOURS = 24;
```
trevorrea: Make this hard coded value configurable.

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -0,0 +1,270 @@
+const NodeClam = require('clamscan');
+const { resolve } = require('path');
+const AWS = require('aws-sdk');
+const s3 = new AWS.S3({ region: 'eu-west-2' });
+
+const deleteFileFromStorage = require('./deleteFileFromStorage');
+
+const TWO_HUNDRED_MEGABYTES = 200 * 1_000_000;
```
trevorrea: Make this hard coded value configurable.

api/controllers/DashboardController.js
```diff
@@ -16,275 +19,400 @@ var dashboardController = {
...
+            },
+            default: {
+                text: casebookStatus,
+                colorClass: 'govuk-tag--blue',
+            },
+        };
+
+        if (!casebookStatus) {
+            return {
+                text: 'Not avialable',
```
simonmoore100: Available is spelt incorrectly

Screenshot 2021-10-04 at 13 57 03

api/controllers/DashboardController.js
```diff
@@ -16,275 +19,400 @@ var dashboardController = {
      * @param res
      *
      */
-    dashboard: function (req, res) {
-        if (!(req.session.passport && req.session.passport.user)){
-            return res.redirect(sails.config.customURLs.userServiceURL + '/sign-in');
+    dashboard(req, res) {
```
simonmoore100, line 22: I've noticed that if casebook is down or returns a 500 error, the dashboard no longer renders. A Page not found error is shown instead.
Can you test this simulating when casebook is down? It should still render the dashboard.

api/controllers/DashboardController.js
simonmoore100: Available isn't spelt correctly

tests/specs/controllers/Dashboard.test.js
```diff
@@ -5,38 +5,313 @@
...
+                totalPages: 2,
+                paginationMessage:
+                    'Showing 1 &ndash; 20 of 35 applications submitted in the last 60 days',
+            };
+            expect(paginationAndPageTotal).to.deep.equal(expectedResult);
         });
     });
 
+    describe('_userFriendlyStatuses()', () => {
+        it('should return Not avialable if casebook does not return a status', () => {
```
simonmoore100: Available isn't spelt correctly

tests/specs/controllers/Dashboard.test.js
```diff
@@ -5,38 +5,313 @@
...
+            // when
+            const returnedValue = dashboardController._userFriendlyStatuses(
+                null,
+                null
+            );
+
+            // then
+            expect(returnedValue).to.deep.equal({
+                text: 'Not avialable',
```
simonmoore100: Available isn't spelt correctly

api/helpers/uploadFileToStorage.js
```diff
@@ -0,0 +1,61 @@
+const multerS3 = require('multer-s3');
+const multer = require('multer');
+const AWS = require('aws-sdk');
+const s3 = new AWS.S3({ region: 'eu-west-2' });
+
+const inDevEnvironment = process.env.NODE_ENV === 'development';
```
simonmoore100, line 6: Rather than all the checks to see if you are in dev mode, just use localstack to create an actual bucket to simulate uploads

api/controllers/EAppSubmittedController.js
```diff
@@ -0,0 +1,180 @@
+const sails = require('sails');
+const AWS = require('aws-sdk');
+const s3 = new AWS.S3({ region: 'eu-west-2' });
```
trevorrea: Do not hard code the region in everywhere. The docs say :-

Using an Environment Variable

You can set the Region using the AWS_REGION environment variable. If you define this variable, the SDK for JavaScript reads it and uses it.

So if we set AWS_REGION in the task definitions then we don't need to set the region anywhere in the code at all

See https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html

I've added this for you at https://github.com/UKForeignOffice/fco-leg-aws/blob/demo/components/leg/ecs_task_definition_application.tf#L10

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -0,0 +1,210 @@
+const sails = require('sails');
+
+const COST_PER_DOCUMENT = 30;
```
simonmoore100: Maybe move this into the .env as it's easier if the price were to change.

api/controllers/ApplicationTypeController.js
RichardBray: At the moment we haven't added the drop-off service to the new applications screen. I think it's something we will add after the MVP but for now it's just the three services.

Screenshot 2021-10-04 at 14 41 51

views/eApostilles/applicationType.ejs
```diff
@@ -0,0 +1,150 @@
...
+                            </table>
+                            <h2 class="heading-medium govuk-!-margin-top-0">If you do not want to pay courier fees</h2>
+                            <p>
+                                Your documents will be returned by courier but if you’re in the UK you can have them returned by post. You must include
+                                an A4-sized envelope with the return address written on it. Attach enough UK postage stamps to cover the return postage.
+                            </p>
+                        </div>
+                    </details>
+                </div>
+                <%if(user_data.user && !user_data.user.premiumEnabled){%>
```
simonmoore100: This logic means the premium journey is shown for people who don't have the flag enabled in the database.

Change it to:

<%if(user_data.user && user_data.user.premiumEnabled){%>

views/eApostilles/applicationType.ejs
RichardBray: With the e-App flag only premium and not drop off, but the MVP will only go out to a handful of notaries before we make more changes and open it up to more notaries so the drop-off might be added later on.

api/controllers/DashboardController.js
```diff
@@ -16,275 +19,400 @@ var dashboardController = {
...
+            pageAttributes.layout = null;
+        }
 
+        return res.view(view, pageAttributes);
+    },
 
-                            return res.view(view,attributes);
-                        }
-                    );
+    _paginationAndPageTotal(results, offset, pageSize) {
```
simonmoore100, line 363: The pagination is incorrect when showing a mixture of e-Apps and regular applications.

Screenshot 2021-10-04 at 15 21 35

api/controllers/DashboardController.js
```diff
@@ -16,275 +19,400 @@ var dashboardController = {
...
+                colorClass: '', // dark blue
+            },
+            default: {
+                text: 'In progress',
+                colorClass: 'govuk-tag--blue',
+            },
+        };
+
+        const standardStatuses = {
+            Despatched: {
```
simonmoore100, line 324: 'Dispatched' is probably the more common way to spell it. Unless this has been passed down from CASEBOOK?

api/helpers/uploadedFileErrorChecks.js
RichardBray: Very true, I honestly don't know why I did it like this 🤦

api/controllers/DashboardController.js
RichardBray: Lol, can't believe no one else picked this up 🤦

api/controllers/DashboardController.js
RichardBray: Fair point. This is a similar thing or a similar fix to the comment you raised here.
https://github.com/UKForeignOffice/loi-application-service/pull/286#discussion_r721218250

It should be sorted now, please pull the code again and let me know if you have any issues.

tests/specs/controllers/Dashboard.test.js
RichardBray: 🤦

api/helpers/uploadFileToStorage.js
RichardBray: This is a good point. We had a discussion about this at CYB before this feature was implemented and we opted to upload files locally without localstack because it was easier. So Jon (don't know if you remember him) implemented local uploads before I picked the project up. 

I'll have a chat with the others and get their thoughts on this but in all honesty, this will have to be done post MVP unless you think it is essential?

api/controllers/DashboardController.js
RichardBray: Yeah this value is passed down from Casebook. I think that's just they way the prefer to spell it 🤷 
Screenshot 2021-10-04 at 18 02 16

api/controllers/EAppSubmittedController.js
RichardBray: This is really helpful. Thanks for sharing this with me 👍

api/controllers/DashboardController.js
RichardBray: This is super odd. I am able to get the pagination working locally on my machine with a mixture of both app types

https://user-images.githubusercontent.com/1377253/135899470-0008dd59-883f-47f1-a61c-a8f038abf7d9.mp4

Would you like to have a chat about this tomorrow? Maybe there's a chance a stored function/procedure hasn't been installed or something 🤔

views/eApostilles/useStandardService.ejs
simonmoore100: I'm getting an error now when hitting this page. I think its because user_data is not being passed in to determine if the user is logged in or not.
Screenshot 2021-10-05 at 10 04 13

config/routes.js
```diff
@@ -1,221 +1,362 @@
 module.exports.routes = {
-
-    'get /healthcheck'                             : 'PublicController.healthcheck',
```
simonmoore100: Could you revert this file back to the way it was? There's so many formatting changes its hard to see what has been added. I think the previous formatting was easier to read.

api/controllers/EAppSubmittedController.js
```diff
@@ -0,0 +1,180 @@
...
+                fileCountError: false,
+            }
+        };
+        req.session.eApp = newSessionData;
+    },
+
+    _sendConfirmationEmail(userDetails, applicationId, req) {
+        const emailAddress = userDetails.email;
+        const applicationRef = applicationId;
+        const { domain } = req._sails.config.session;
```
simonmoore100: Take a look at how we do this in the notification service as we construct URLs for other emails and there is already a variable for this. It'll be better than working out the domain from the redis session.

https://github.com/UKForeignOffice/loi-notification-service/blob/develop/app/routes.js#L242

api/controllers/FileUploadController.js
```diff
@@ -0,0 +1,133 @@
...
+const uploadFileToStorage = require('../helpers/uploadFileToStorage');
+const deleteFileFromStorage = require('../helpers/deleteFileFromStorage');
+const {
+    virusScanFile,
+    checkTypeSizeAndDuplication,
+    displayErrorAndRemoveLargeFiles,
+    connectToClamAV,
+} = require('../helpers/uploadedFileErrorChecks');
+
+const MAX_FILES = 50;
```
simonmoore100: Again, maybe make this an env variable. We occasionally get requests to limit applications if the office are very busy so any way to make this kind of thing configurable is good.

api/controllers/OpenEAppController.js
```diff
@@ -0,0 +1,173 @@
+const sails = require('sails');
+const request = require('request-promise');
+const crypto = require('crypto');
+const dayjs = require('dayjs');
+const duration = require('dayjs/plugin/duration');
+dayjs.extend(duration);
+
+const MAX_DAYS_TO_DOWNLOAD = 21;
```
simonmoore100: Again, convert to env variable

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -0,0 +1,271 @@
...
+    } else {
+        req.session.eApp.uploadMessages.errors.push({
+            filename: file.originalname,
+            errors,
+        });
+    }
+}
+
+function displayErrorAndRemoveLargeFiles(req) {
+    const TWO_HUNDRED_MEGABYTES =
```
simonmoore100: What if you change the 'file_upload_size_limit' value from 200 to 300? :)

Maybe change the name of the const TWO_HUNDRED_MEGABYTES to be more general

views/eApostilles/applicationType.ejs
simonmoore100: Speaking with James Richards today on our standup, they are expecting the drop off service to be there even for the beta so it looks like it needs to be enabled. Can you re-confirm this requirement?

api/controllers/EAppSubmittedController.js
RichardBray: Hey Simon, yeah the code looks a bit confusing but the domain is actually retrieved from the .env file but it happens to be part of the session object if that makes sense.

It looks like the notifications service is doing a similar thing but in a much more convoluted way. 
notify.js > common.js > server.js > routes.js

But that refactor is for a future PR 😄

simonmoore100: Yes I see you are getting it from the redis session var, but now we have 1 email constructed differently than the rest of the emails in the service. 

Please change it to be constructed the same as the others in the notification service. It just makes it easier for future troubleshooting/development.

api/controllers/DashboardController.js
simonmoore100: As per our meeting this morning, it looks like this is a genuine bug so Richard is going to have a look.

RichardBray: Hey Simon this has been fixed now. Man, I dislike stored procedures. If you reapply the updated function in EIA-101_update_dashboard_function.sql it should work 🤞

api/controllers/EAppSubmittedController.js
RichardBray: This should be done now. Please let me know if you think anything else should be changed.

views/eApostilles/useStandardService.ejs
simonmoore100: Not sure if you saw the above comment @RichardBray

simonmoore100 (approved):
Happy to merge. Everything has been addressed bar the exit page showing an error. This minor issue can be addressed in the coming days before release.

PR #289, opened 2021-10-11
Bug/EIA-199 Fix issue with accessing other user's eApp data

RichardBray:
https://jira.digital.csd.fcdo.gov.uk/browse/EIA-199


simonmoore100 (approved):
Tested locally and it's successfully blocking the CASEBOOK retrieval for an e-App that isn't yours.

PR #291, opened 2021-10-13
Bug/EIA-202 Increase security of file type check

RichardBray:
https://jira.digital.csd.fcdo.gov.uk/browse/EIA-201

### How to tests
- rename the extention of any file to .pdf
- upload the file
- it should prevent you from doing so

Screenshot 2021-10-13 at 11 33 41


simonmoore100 (approved):
Apart from the title of the ticket being wrong, the functionality seems to block unwanted files from being uploaded.

RichardBray:
> Apart from the title of the ticket being wrong, the functionality seems to block unwanted files from being uploaded.

Ha yeah, I don't know why I did that. I blame it on kids and not getting enough sleep 😬

PR #292, opened 2021-10-14
Bug/EIA-198 Refactor and test dashboard code

RichardBray:
# Description

This is primarily a PR to understand and refactor an already existing controller. Kind of leaving the code better than you found it but making the code that already existed better instead of the code that you added.

https://jira.digital.csd.fcdo.gov.uk/browse/EIA-198

## Type of change

# How Has This Been Tested?

- [x] Manual end-to-end
- [x] Unit tests

# Checklist:

- [x] Any dependent changes have been merged and published in downstream modules


.gitignore
```diff
@@ -146,10 +146,6 @@ coverage
 # Compiled binary addons (http://nodejs.org/api/addons.html)
 build/Release
 
-# Dependency directory
-# https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
-node_modules
-
```
RichardBray, line 152: There were two "node_modules" in this file so I just removed one.

api/controllers/DashboardController.js
```diff
@@ -87,6 +53,50 @@ var dashboardController = {
         });
     },
 
+    _calculateSortParams(req, res, userData, totalApplications) {
+        const pageSize = 20;
+        const currentPage = req.query.page || 1;
+        const offset = pageSize * (currentPage - 1);
+        const sortOrder = Number(req.query.sortOrder) || -1;
```
RichardBray, line 58: You might miss this but this line casts a string to a number to make `secondarySortOrder` and `secondaryDirection` work properly. It's now comparing a number with a number instead of a string with a number.

api/controllers/DashboardController.js
```diff
@@ -122,7 +132,104 @@ var dashboardController = {
...
+            json: true,
+            useQuerystring: true,
+            qs: leg_app_stat_struc,
+        };
+
+        return request(options)
+            .then((response) => {
+                const isErrorResponse =
+                    typeof response === 'object' &&
+                    response.hasOwnProperty('errors');
```
jenbutongit: when would response not be an object? could it just check for response.error?

api/controllers/DashboardController.js
```diff
@@ -133,177 +240,66 @@ var dashboardController = {
...
-
-                return dashboardController._redirectToPage(
-                    pageAttributes,
-                    req,
-                    res
+                result.app_status = dashboardController._userFriendlyStatuses(
+                    appStatus,
+                    result.applicationtype
                 );
+                result.tracking_ref = trackRef.hasOwnProperty(uniqueAppId)
```
jenbutongit: are you checking for `null` specifically? i.e. `if(result.tracking_ref === null)`? If not I think `result.tracking_ref = trackRef[uniqueAppId]` would suffice. It will give you undefined but it would take out one more logic check out (trackRef.hasOwnProperty).

api/controllers/DashboardController.js
```diff
@@ -87,6 +53,50 @@ var dashboardController = {
...
+        const secondaryDirection =
+            (sortOrder === 1 || sortOrder === -1) ? null : 'desc';
+        const storedProcedureArgs = {
+            replacements: {
+                userId: req.session.passport.user,
+                pageSize,
+                offset,
+                sortOrder: Math.abs(sortOrder).toString(),
+                direction,
+                queryString: '%' + searchCriteria + '%',
```
jenbutongit, line 81: should this be using `querystring.encode` or `encodeURIComponent()`? (These are both native node or js so no need to worry about another library), generally they are much safer and better tested than DIY-encoding

api/controllers/DashboardController.js
```diff
@@ -87,6 +53,50 @@ var dashboardController = {
...
+        const direction = isNumberPositive ? 'asc' : 'desc';
+        const searchCriteria =
+            req.allParams().dashboardFilter || req.query.searchText || '';
+        //If user has specifically selected to filter by date (sortOrder eq 1 or -1)
+        //then we don't want to secondary sort by date again! But if a user filters by
+        //reference number for example, then secondary sort on the date as well.
+        const secondarySortOrder =
+            (sortOrder === 1 || sortOrder === -1) ? null : '1';
+        const secondaryDirection =
+            (sortOrder === 1 || sortOrder === -1) ? null : 'desc';
```
jenbutongit: it seems like sortOrder is already casted with `|| -1`, so wont these ternaries always be true?

jenbutongit (changes requested):

api/controllers/DashboardController.js
RichardBray: Good point. I've had a 500 from Casebook before that wasn't an object that's why I've put this check-in, but I guess if it's not an object the .catch will pick it up 🤔

api/controllers/DashboardController.js
RichardBray: My worry here is that if one of the values does not exist an error will pop up in the logs saying something like value cannot be found. For some reason, this codebase doesn't allow optional chaining.

What are your thoughts on this instead of the ternery?:
```js
result.tracking_ref = trackRef.hasOwnProperty(uniqueAppId) && trackRef[uniqueAppId];
```

jenbutongit: that error usually occurs when you try to access a property on an undefined value, however you've got `let trackRef = {}`. so `trackRef.anything` will not cause the app to crash, since `trackRef` is not an undefined value. If you tried to chain one more time, `trackRef.somethingUndefined.somethingElse` that will definitely cause an error. 

When you use the `&&` operator, you definitely need to make sure you are not doing a strict check for a null or undefined value with `===` somewhere else in the code. You will end up with either `false` or `trackRef[uniqueAppId]`.

api/controllers/DashboardController.js
RichardBray: I actually have no idea what the benefit of the percentages symbols are. I've done some research and they don't seem to be a wildcard operator for sequelize queries:
https://sequelize.org/master/manual/raw-queries.html

I think they mean if there are values that start or end with the searchCriteria they will be presented to the user.

Maybe I'm missing something but I don't know how useful the encodeURIComponent() function would be here.

api/controllers/DashboardController.js
RichardBray: ah fair point, I forgot about the `trackRef = {}`, yep you're right, with that in place it should be fine. I'll adjust the code accordingly :)

api/controllers/DashboardController.js
RichardBray: It's super weird but you're partly right. When the page first loads the ternaries are true because `req.query.sortOrder` is undefined (line 60). But `req.query.sortOrder` always passes a string of a number, and if that number is not 1 (or -1), then they are false.

api/controllers/DashboardController.js
jenbutongit: AH, my bad, this is a db queryString (not a URL one, so no, my suggestions wont help you). When you do %yourSearchString%, sql will try to search for that substring. If you just have one on either side, '%'+ searchCriteria, it will only search for when searchCriteria occurs at the end of the string. (and vice versa for searchCriteria + '%')

api/controllers/DashboardController.js
```diff
@@ -122,7 +132,102 @@ var dashboardController = {
...
+
+        // First build array of application references to be passed to the Casebook Status API for this page. Can submit up to 20 at a time.
+        const applicationReferences = results.map(
+            (resultItem) => resultItem.unique_app_id
+        );
+
+        // Create Request Structure
+
+        const leg_app_stat_struc = {
+            timestamp: new Date().getTime().toString(),
```
jenbutongit: new Date().getTime().toString() does the same thing as `Date.now()` (you don't need the `new` keyword)

api/controllers/DashboardController.js
jenbutongit: How about something like this 
```suggestion
       let secondarySortOrder, secondaryDirection;
       
       if (sortOrder === 1 || sortOrder === -1) {
       ...
       }
       //or even
       const hasSortOrder = sortOrder === 1 || sortOrder === -1
       if(!hasSortOrder) {..} 
```
your comment already explains that but I usually skip over them (as you can tell from my confusion). 😄

api/controllers/DashboardController.js
```diff
@@ -122,7 +132,102 @@ var dashboardController = {
...
+        };
+
+        const queryStr = apiQueryString.stringify(leg_app_stat_struc);
+
+        let certPath;
+        try {
+            certPath = req._sails.config.casebookCertificate;
+        } catch (err) {
+            sails.log.error('Null certificate path: [%s] ', err);
+            certPath = null;
```
jenbutongit: certPath is already `undefined`, I don't think you have to redefine it as `null` unless you're explicitly checking for null values?

api/controllers/DashboardController.js
```diff
@@ -122,7 +132,102 @@ var dashboardController = {
...
+            keyPath = null;
+        }
+
+        // calculate HMAC string and encode in base64
+
+        const hash = crypto
+            .createHmac('sha512', req._sails.config.hmacKey)
+            .update(Buffer.from(queryStr, 'utf-8'))
+            .digest('hex')
+            .toUpperCase();
```
jenbutongit: have I seen this a few times? would it be worth hoisting this to config or a shared utils file?

```suggestion
        const hashWithHmac = (value) => crypto
            .createHmac('sha512', req._sails.config.hmacKey)
            .update(Buffer.from(value, 'utf-8'))
            .digest('hex')
            .toUpperCase();
```

api/controllers/DashboardController.js
```diff
@@ -122,7 +132,102 @@ var dashboardController = {
...
+            useQuerystring: true,
+            qs: leg_app_stat_struc,
+        };
+
+        return request(options)
+            .then((response) => {
+                const responseHasErrors = response.hasOwnProperty('errors');
+                if (responseHasErrors) {
+                    sails.log.error(
+                        `Invalid response from Casebook Status API call:  ${response.message}`
```
jenbutongit, line 178: it might be worth adding the status code as well

api/controllers/DashboardController.js
```diff
@@ -133,177 +238,62 @@ var dashboardController = {
...
+            dashboardController._paginationAndPageTotal(
+                results,
+                offset,
+                pageSize
+            );
         if (results.length === 0) {
             if (currentPage != 1) {
                 return res.view('404.ejs');
             } else {
                 sails.log.error('No results found.');
```
jenbutongit, line 214: what happens to users who see this? are they meant to see a specific page? looks like they get "stuck" here, as opposed to currentPage != 1 users who get a 404 explicitly returned to them

tests/specs/controllers/Dashboard.test.js
```diff
@@ -47,147 +56,407 @@ describe('DashboardController:', function () {
...
+                session: {
+                    passport: {
+                        user: 123,
+                    },
+                },
+                query: {
+                    page: null,
+                    sortOrder: null,
+                    searchText: null,
+                },
```
jenbutongit: would a user's query string ever be `?page=null&sortOrder=null` etc? does the logic still work when `query : { }`

jenbutongit (changes requested):

api/controllers/DashboardController.js
RichardBray: tbh, I'm not a massive fan of this whole checking if certificate and key already exists. It's not anywhere else in the codebase and if those two are missing from the .env file then the whole app wouldn't work well not just this section.

api/controllers/DashboardController.js
RichardBray: Yeah this is there all over the place, I agree, would make sense to move this repeated code 👍

api/controllers/DashboardController.js
RichardBray: Even after refactoring the code isn't super easy to follow 😞 
After the log is printed the user is redirected to the dashboard page which should show this text where the applications should be:
```
Once you've submitted an application, you'll see details of your application history on this page.
```

If that makes sense. I'll actually test that to make sure that's still the case.

RichardBray: Still works 😄 

Screenshot 2021-10-25 at 12 44 56

tests/specs/controllers/Dashboard.test.js
RichardBray: Good point, no it shouldn't ever be `?page=null&sortOrder=null` and technically the logic should still work with an empty query but I could write a test for that 🤷

api/controllers/DashboardController.js
RichardBray: Makes sense. Although since this controller ends up returning mainly html pages what would the benefit of a status code be?

jenbutongit: it would be for quicker debugging, so if it comes back with 5xx - it's probably something that went wrong with CB, but if it retuned with 4xx there's probably something up with the payload we're sending. if CB error messages are usually helpful enough you can skip out on this

tests/specs/controllers/Dashboard.test.js
```diff
@@ -47,147 +56,407 @@ describe('DashboardController:', function () {
...
+            expectedSortParams.storedProcedureArgs.replacements.secondarySortOrder =
+                '1';
+            expectedSortParams.storedProcedureArgs.replacements.secondaryDirection =
+                'desc';
+            expectedSortParams.displayAppsArgs.sortOrder = -3;
+
+            expect(result).to.deep.equal(expectedSortParams);
+        });
+
+        it('', () => {
```
jenbutongit: missing test name?

api/helpers/prepareApiOptions.js
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @param {{
+ *   url: string,
+ *   req: Request,
+ *   apiOptions: Object,
+ *   useApiQueryString: boolean,
+ *   refParam: string,
+ * }} kwargs
+ */
+function prepareAPIOptions(kwargs) {
```
jenbutongit: kwargs? 🤔

jenbutongit: How about default requests? Given you're still using the `request` lib, you can use the convenience defaults method. 

``` js
// baseRequest.js
// const hash, key, cert
const baseOptions = {
 agentOptions: {
             cert,
             key,
         },
         headers: {
             hash,
         },
}
export const baseRequest = request.defaults(defaultOptions)
```

```.js
// whereverImMakingTheRequest.js
const { baseRequest } = require('baseRequest')
const myAdditionalOptions = {...}
baseRequest(myAdditionalOptions, (req) => { //this call should contain the correct agentOptions and headers
})
```

jenbutongit (changes requested):

api/helpers/prepareApiOptions.js
RichardBray: Lol `kwargs` is more of a python thing, it stands for keyword arguments, couldn't think of a better name for it.

tests/specs/controllers/Dashboard.test.js
RichardBray: 🤦 don't know how that slipped through

api/helpers/prepareApiOptions.js
```diff
@@ -0,0 +1,50 @@
...
+const apiQueryString = require('querystring');
+
+function prepareAPIOptions(prepareOptionsArgs) {
+    const {
+        url:uri,
+        req,
+        apiOptions,
+        useApiQueryString = false,
+        refParam,
+    } = prepareOptionsArgs;
```
jenbutongit: If you don't go for a service pattern, let's match the request options as closely as possible. There's no need to rename url to uri, and could just cause confusion. 

```.js
function prepareAPIOptions(prepareOptionsArgs) {
    const {
        uri,  // <- rename to just uri. It's also not used apart from the end, so let's remove it from here
        req, 
        apiOptions, // <- this is never transformed or used apart from the return
        useApiQueryString = false, 
        refParam // <- reference parameter? query would just be clearer. a parameter could be `/{id}/{file}`, or query parameter. 
    } = prepareOptionsArgs;
```

```.js
function prepareAPIOptions(prepareOptionsArgs) {
    const {
        req, 
        useApiQueryString = false, 
       query,
        ...rest, // <- rest will be options that can be directly passed into request(options)
    } = prepareOptionsArgs;
    
  const queryParameters = { timestamp, ...query }
  let queryString 
  // if/else block can stay the same
  
  // const hash.. 
  
  return {
  ...rest, (should contain { uri: '..' },
  headers,
  agentOptions,
  //etc
  } 
```
because this uses `...rest`, you don't need to pass in APIOptions as an object. 

So your calls would go from this 
```.js
const myOptions = prepareAPIOptions({ ApiOptions: { body: 'some data to send' }, url: config.someUrl })

// to
const myOptions = prepareAPIOptions({body: 'some data to send' , uri: config.someUrl })
// If I do not need to "prepare" my request, { body: 'some data to send',  uri: config.someUrl } would still be valid for request.post(). 

```

jenbutongit (approved):
LGTM. We can go through services and DI when you're next in the office if you like.. it will be easier to explain in person 😄

RichardBray:
> LGTM. We can go through services and DI when you're next in the office if you like.. it will be easier to explain in person 😄

Sure thing. Happy to have a call about it as well if you have time.

PR #300, opened 2021-10-28
Feat/eia 208 accessibility testing

RichardBray:
https://jira.digital.csd.fcdo.gov.uk/browse/EIA-208


cypress/integration/a11y.spec.js
```diff
@@ -0,0 +1,163 @@
...
+    context('Pre login pages', () => {
+        beforeEach(() => {
+            cy.visit('/');
+        });
+        it('Get your document legalised', () => {
+            checkA11y();
+        });
+
+        it('Choose a service', () => {
+            cy.get('.govuk-button--start').contains('Start now').click();
```
jenbutongit: I would recommend changing these matchers. Class matching is quite flakey in general. 

I'm using https://testing-library.com/docs/cypress-testing-library/intro/. You should be familiar with it from react-testing-library. I try to exclusively use `findByLabelText` or `findByRole`. It gives you an added layer of accessibility confidence, since you are finding/getting by what is accessible to screenreaders. https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-byrole-most-of-the-time

jenbutongit (changes requested):

cypress/integration/a11y.spec.js
```diff
@@ -0,0 +1,187 @@
...
+            checkA11y();
+        });
+    });
+
+    describe('Post login', () => {
+        function uploadTestFile() {
+            cy.get('.multi-file-upload__input').attachFile('test.pdf');
+            cy.get('.govuk-button--secondary').contains('Upload').click();
+            cy.wait(500);
```
jenbutongit: cypress waits on its own, adding a hardcoded wait will contribute to flakiness. Is there an element that pops up? you can do `cy.get('.component-that-hasnt-appeared-yet', {timeout: ..})` if the default timeout isn't long enough.

cypress/integration/a11y.spec.js
```diff
@@ -0,0 +1,187 @@
...
+
+        function clickContinueBtn() {
+            cy.get('.column-two-thirds .govuk-button')
+                .contains('Continue')
+                .click();
+        }
+
+        function checkFirstRadioAndClickContinue() {
+            cy.get('[type="radio"].govuk-radios__input').first().check();
+            clickContinueBtn();
```
jenbutongit: I know that these are just a11y tests, but I have no idea what the first radio is, and what it does, where it goes etc. 

With testing library, it would look like this 
```suggestion
            const { getByLabelText, getByRole } = cy;
            getByLabelText('First Item').click()
            getByRole('button').click()
```
And then I would stop nesting it inside a function, since it is just as readable as "checkFirstRadioAndClickContinue" (if not more), and is more flexible.

cypress/integration/a11y.spec.js
```diff
@@ -0,0 +1,187 @@
...
+
+            it('Check if the documents are eligible for the e-Apostille service', () => {
+                checkFirstRadioAndClickContinue();
+                checkFirstRadioAndClickContinue();
+                checkA11y();
+            });
+
+            it('Have the PDFs been notarised and digitally signed by a notary?', () => {
+                for (var i = 0; i < TOTAL_ELIBIGILITY_QUESTIONS; ++i) {
+                    checkFirstRadioAndClickContinue();
```
jenbutongit: what about the other routes?

docs/automated-a11y-testing.md
```diff
@@ -0,0 +1,38 @@
...
+
+2 - The tests will all fail if the Cognito Login screen is not bypassed. Please make sure you have your IP whitelisted to avoid this.
+
+3 - Don't run tests on localhost! For some reaons testing on localhost takes a long time for the page to load. THe *cypress.json* file has the demo site at the baseUrl but feel free to change that if necessary.
+```json
+{
+  "baseUrl": "https://www.demo.legalisation.fcodev.org.uk",
+  ...
+}
+```
```
jenbutongit: did you mean to keep ... here?

this should resolve the syntax highlighting issue
```suggestion
```json5
{
  "baseUrl": "https://www.demo.legalisation.fcodev.org.uk",
  // ...
}
```
```

File: views/eApostilles/checkUploadedDocuments.ejs [hunk 6]
```diff
@@ -46,7 +46,7 @@
         </dl>
     <% } %>
 
-    <h3 class="heading-medium govuk-!-margin-top-2">Cost</h3>
+    <h2 class="heading-medium govuk-!-margin-top-2">Cost</h2>
```
jenbutongit, line 49: it looks like there are upto 3 `h2`s on this page now. Is that right?

views/eApostilles/openEApp.ejs
```diff
@@ -60,7 +60,7 @@
                         data privacy reasons. You will need to pay for another application to have the documents legalised again.
                     </strong>
                 </div>
-                <h3 class="heading-medium">Download documents</h3>
+                <h2 class="heading-medium">Download documents</h2>
```
jenbutongit, line 63: there are quite a few on this page too. Is this semantically correct?

jenbutongit (changes requested):

cypress/integration/a11y.spec.js
RichardBray: Good shout. I didn't feel the need to test them since they are just static exit pages but I can do them as well.

docs/automated-a11y-testing.md
RichardBray: This is a neat trick. json5 looks like a more advanced version of jsonc. Have you used it in any projects before?

views/eApostilles/checkUploadedDocuments.ejs
RichardBray: Yep, that's correct. I was under the assumption you could only have one h1 but multiple h2-6's on a page. Maybe it's different for GOV pages?

cypress/integration/a11y.spec.js
RichardBray: to be fair I didn't test it without the wait, I added it because I assumed it was necessary to wait for the file to actually upload but looks like it works without it 👍

views/eApostilles/checkUploadedDocuments.ejs
jenbutongit: that's right. If all 3 headings/sections are as important as each other that's fine. Some users using screen readers like to navigate by headings so it's just good to keep in mind

docs/automated-a11y-testing.md
jenbutongit: Not really, I think I always end up using transpiling tools that will get rid of json5 (or js/ts) trailing object commas, but I know it exists (and GitHub will give you nicer parsing if you have comments etc)

cypress/integration/a11y.spec.js
```diff
@@ -0,0 +1,200 @@
...
+
+            it('You cannot use this service', () => {
+                checkRadioAndClickContinue('e-Apostille service');
+                checkRadioAndClickContinue('No');
+                checkA11y();
+            });
+
+            it('Have the PDFs been notarised and digitally signed by a notary?', () => {
+                checkRadioAndClickContinue('e-Apostille service');
+                for (var i = 0; i < TOTAL_ELIBIGILITY_QUESTIONS - 1; ++i) {
```
jenbutongit: this can start on `i = 1` so there's no need to `-1` from eligibility Qs

cypress/integration/a11y.spec.js
```diff
@@ -0,0 +1,200 @@
+describe('Check accessiblity', () => {
+    const TOTAL_ELIBIGILITY_QUESTIONS = 3;
+
```
jenbutongit, line 5: you can destructure up here all of the methods that you're using frequently like 
`const { findByLabelText, findByRole, setCookie } = cy` to improve the readability

jenbutongit (changes requested):

jenbutongit (approved):

PR #305, opened 2021-11-10
Feat/eia 190 replace request

RichardBray:
# Description
Replace depreciated request library with [Axios](https://github.com/axios/axios). This also removes the legacy [querystring](https://nodejs.org/api/querystring.html) Node API in favour of the newer [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) one.

https://jira.digital.csd.fcdo.gov.uk/browse/EIA-190

## Type of change

Please delete options that are not relevant.

- [x] Remove depreciated library

# How Has This Been Tested?

- [x] Manually
- [x] Unit tests

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/api/controllers/DashboardController.js
+++ b/api/controllers/DashboardController.js
@@ -144,45 +144,30 @@ const dashboardController = {
                 sails.log.error('No results found.');
             }
         }
-        const {data: apiResponse} = await dashboardController._getDataFromCasebook(
-            req,
-            results
-        );
+        const { data: apiResponse } =
+            await dashboardController._getDataFromCasebook(req, results);
+
         return dashboardController._addCasebookStatusesToResults(apiResponse, {
             ...displayAppsArgs,
             results,
         });
     },
 
-    _getDataFromCasebook(req, results) {
+    async _getDataFromCasebook(req, results) {
         const applicationReferences = results.map(
             (resultItem) => resultItem.unique_app_id
         );
-
-        const url = req._sails.config.customURLs.applicationStatusAPIURL;
         const queryParamsObj = {
             timestamp: Date.now().toString(),
             applicationReference: applicationReferences,
         };
 
-        return CasebookService.get(url, {
-            params: queryParamsObj,
-        })
-            .then((response) => {
-                const responseHasErrors = response.hasOwnProperty('errors');
-                if (responseHasErrors) {
-                    sails.log.error(
-                        `Invalid response from Casebook Status API call:  ${response.message}`
-                    );
-                    return response.status(500);
-                }
-                return response;
-            })
-            .catch((err) => {
-                sails.log.error(
-                    `Error returned from Casebook API call: ${err}`
-                );
-            });
+        try {
+            return await CasebookService.getApplicationStatus(queryParamsObj);
+        } catch (error) {
+            sails.log.error(error);
+            return res.serverError();
+        }
     },
 
     _addCasebookStatusesToResults(apiResponse, displayAppsArgs) {
--- a/api/controllers/DashboardController.js
+++ b/api/controllers/DashboardController.js
@@ -99,7 +99,6 @@ const DashboardController = {
     },
 
     _getApplications(storedProcedureArgs, displayAppsArgs, electronicEnabled) {
-        const { res } = displayAppsArgs;
         const applicationType = electronicEnabled
             ? 'electronic and paper'
             : 'paper';
@@ -118,8 +117,7 @@ const DashboardController = {
                 );
             })
             .catch((error) => {
-                sails.log.error(error);
-                return res.serverError();
+                throw new Error(error);
             });
     },
 
@@ -143,8 +141,7 @@ const DashboardController = {
                 sails.log.error('No results found.');
             }
         }
-        const apiResponse = await DashboardController._getDataFromCasebook(
-            req,
+        const {data: apiResponse} = await DashboardController._getDataFromCasebook(
             results
         );
         return DashboardController._addCasebookStatusesToApplicationRow(
@@ -158,16 +155,9 @@ const DashboardController = {
 
     async _getDataFromCasebook(results) {
         try {
-            const applicationReferences = results.map(
-                (resultItem) => resultItem.unique_app_id
-            );
-
-            return await CasebookService.getApplicationStatus(
-                applicationReferences
-            );
+            return await CasebookService.getApplicationsStatuses(results);
         } catch (error) {
-            sails.log.error(error);
-            return res.serverError();
+            throw new Error(error);
         }
     },
 
--- a/api/controllers/DashboardController.js
+++ b/api/controllers/DashboardController.js
@@ -159,17 +159,15 @@ const dashboardController = {
             (resultItem) => resultItem.unique_app_id
         );
 
+        const url = req._sails.config.customURLs.applicationStatusAPIURL;
         const queryParamsObj = {
             timestamp: Date.now().toString(),
             applicationReference: applicationReferences,
         };
 
         return CasebookService.get({
-            uri: req._sails.config.customURLs.applicationStatusAPIURL,
-            json: true,
-            useQuerystring: true,
-            promise: true,
-            qs: queryParamsObj,
+            url,
+            params: queryParamsObj,
         })
             .then((response) => {
                 const responseHasErrors = response.hasOwnProperty('errors');
--- a/api/controllers/DashboardController.js
+++ b/api/controllers/DashboardController.js
@@ -165,8 +165,7 @@ const dashboardController = {
             applicationReference: applicationReferences,
         };
 
-        return CasebookService.get({
-            url,
+        return CasebookService.get(url, {
             params: queryParamsObj,
         })
             .then((response) => {
--- a/api/controllers/DashboardController.js
+++ b/api/controllers/DashboardController.js
@@ -145,7 +145,7 @@ const dashboardController = {
             }
         }
         const { data: apiResponse } =
-            await dashboardController._getDataFromCasebook(req, results);
+            await dashboardController._getDataFromCasebook(results);
 
         return dashboardController._addCasebookStatusesToResults(apiResponse, {
             ...displayAppsArgs,
@@ -153,16 +153,16 @@ const dashboardController = {
         });
     },
 
-    async _getDataFromCasebook(req, results) {
-        const applicationReferences = results.map(
-            (resultItem) => resultItem.unique_app_id
-        );
-        const queryParamsObj = {
-            timestamp: Date.now().toString(),
-            applicationReference: applicationReferences,
-        };
-
+    async _getDataFromCasebook(results) {
         try {
+            const applicationReferences = results.map(
+                (resultItem) => resultItem.unique_app_id
+            );
+            const queryParamsObj = {
+                timestamp: Date.now().toString(),
+                applicationReference: applicationReferences,
+            };
+
             return await CasebookService.getApplicationStatus(queryParamsObj);
         } catch (error) {
             sails.log.error(error);
--- a/api/controllers/DashboardController.js
+++ b/api/controllers/DashboardController.js
@@ -158,12 +158,10 @@ const dashboardController = {
             const applicationReferences = results.map(
                 (resultItem) => resultItem.unique_app_id
             );
-            const queryParamsObj = {
-                timestamp: Date.now().toString(),
-                applicationReference: applicationReferences,
-            };
 
-            return await CasebookService.getApplicationStatus(queryParamsObj);
+            return await CasebookService.getApplicationStatus(
+                applicationReferences
+            );
         } catch (error) {
             sails.log.error(error);
             return res.serverError();
--- a/api/controllers/DashboardController.js
+++ b/api/controllers/DashboardController.js
@@ -144,7 +144,7 @@ const dashboardController = {
                 sails.log.error('No results found.');
             }
         }
-        const apiResponse = await dashboardController._getDataFromCasebook(
+        const {data: apiResponse} = await dashboardController._getDataFromCasebook(
             req,
             results
         );
--- a/api/controllers/FileDownloadController.js
+++ b/api/controllers/FileDownloadController.js
@@ -1,4 +1,6 @@
 const sails = require('sails');
+const stream = require('stream');
+const util = require('util');
 const CasebookService = require('../services/CasebookService');
 
 const FileDownloadController = {
@@ -44,15 +46,12 @@ const FileDownloadController = {
 
     _apostilleRefBelongToApplication(req, res) {
         FileDownloadController._urlErrorChecks(req, res);
-        const url = req._sails.config.customURLs.applicationStatusAPIURL;
         const queryParamsObj = {
             timestamp: Date.now().toString(),
             applicationReference: req.params.unique_app_id,
         };
 
-        return CasebookService.get(url, {
-            params: queryParamsObj,
-        })
+        return CasebookService.getApplicationStatus(queryParamsObj)
             .then((response) => {
                 const appInfo = response.data[0];
                 if (appInfo) {
@@ -90,34 +89,23 @@ const FileDownloadController = {
     },
 
     _streamFileToClient(req, res) {
-        const url = req._sails.config.customURLs.apostilleDownloadAPIURL;
+        const streamFinished = util.promisify(stream.finished);
+
         const queryParamsObj = {
             timestamp: Date.now().toString(),
             apostilleReference: req.params.apostilleRef,
         };
 
         sails.log.info('Downloading file from Casebook');
-        CasebookService.get(url, {
-            responseType: 'stream',
-            params: queryParamsObj,
-        })
-            .then(function (response) {
-                if (response.status !== 200) {
-                    sails.log.error(`Casebook returned ${response.statusCode}`);
-                } else {
-                    response.data.pipe(res);
-                    sails.log.info('File successfully downloaded');
-                }
+
+        CasebookService.getApostilleDownload(queryParamsObj)
+            .then((response) => {
+                response.data.pipe(res);
+                return streamFinished(res);
             })
             .catch((err) => {
                 throw new Error(err);
             });
-    },
-
-    _renamePDFFromHeader(req, res) {
-        res.headers[
-            'content-disposition'
-        ] = `attachment; filename=LegalisedDocument-${req.params.apostilleRef}.pdf`;
     },
 };
 
--- a/api/controllers/FileDownloadController.js
+++ b/api/controllers/FileDownloadController.js
@@ -44,20 +44,17 @@ const FileDownloadController = {
 
     _apostilleRefBelongToApplication(req, res) {
         FileDownloadController._urlErrorChecks(req, res);
-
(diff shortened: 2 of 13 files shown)


api/controllers/DashboardController.js
```diff
@@ -159,17 +159,14 @@ const dashboardController = {
...
-        return CasebookService.get({
-            uri: req._sails.config.customURLs.applicationStatusAPIURL,
-            json: true,
-            useQuerystring: true,
-            promise: true,
-            qs: queryParamsObj,
+        return CasebookService.get(url, {
+            params: queryParamsObj,
         })
             .then((response) => {
```
jenbutongit: you can use async/await with axios. Generally I prefer it since it removes a layer of nesting generally makes the code more readable

api/controllers/DashboardController.js
```diff
@@ -159,17 +159,14 @@ const dashboardController = {
             (resultItem) => resultItem.unique_app_id
         );
 
+        const url = req._sails.config.customURLs.applicationStatusAPIURL;
```
jenbutongit: it looks like this url is being used a few times in different controllers. I'd recommend adding this to casebookService, and removing `_getDataFromCasebook` 

DashboardController, FileUploadController doesn't need to be concerned on how and where to to fetch application status, it can just "ask" casebookService to... serve it to them `casebookService.getApplicationStatus(..)`

It also means that these controllers don't actually have to do any logging when a casebook calls fails - casebookService can do that. DashboardController just has to log anything specific to dashboardController

api/services/CasebookService.js
```diff
@@ -1,53 +1,65 @@
...
 
-    _returnRequestMethods(method, options) {
-        const requestType = options.promise ? requestPromise : request;
-        delete options.promise;
+function queryParamObjToStr(queryParamsObj) {
+    const params = new URLSearchParams();
 
-        const authParams = CasebookService._createAuthParams(options.qs);
-        const optionsWithAuthParams = { ...options, ...authParams };
+    Object.entries(queryParamsObj).forEach(([key, value]) => {
```
jenbutongit: any reason why this is DIY? node.js has a way of handling objects and query strings built in https://nodejs.org/api/querystring.html#querystringstringifyobj-sep-eq-options

api/services/CasebookService.js
```diff
@@ -1,53 +1,65 @@
...
-    },
-};
+function addHmacToQueryParam(data, _headers) {
+    data = {
+        hmac: hmacKey
+    };
+    return JSON.stringify(data);
+}
+
+CasebookService.interceptors.request.use(addHashToHeader);
```
jenbutongit: should this be done as part of the request, rather than basically after the request has been made via intercepting?

api/controllers/OpenEAppController.js
```diff
@@ -64,44 +62,24 @@ const OpenEAppController = {
...
-            headers: {
-                hash,
-                'Content-Type': 'application/json; charset=utf-8',
-            },
-            json: true,
-            qs: queryParamsObj,
-        };
 
-        return request(options)
+        return CasebookService.get(url, {
```
jenbutongit: maybe `casebook.getApplicationDataFromCasebook`?

api/controllers/OpenEAppController.js
```diff
@@ -64,44 +62,24 @@ const OpenEAppController = {
...
-                    response.hasOwnProperty('errors');
+                    typeof response.data === 'object' &&
+                    response.data.hasOwnProperty('errors');
                 if (isErrorResponse) {
-                    sails.log.error(response.message);
+                    sails.log.error(response.data.message);
                     return res.serverError();
                 }
-                return response;
+                return response.data;
```
jenbutongit: would it just be easier to check for response code?

jenbutongit (changes requested):

api/services/CasebookService.js
RichardBray: Good point. The service was using queryString before but apparently, it's depreciated so I got rid of it. And it looks like `URLSearchParams` doesn't have a native way of parsing arrays.

Screenshot 2021-11-11 at 10 05 40

api/services/CasebookService.js
RichardBray: That would be ideal but I couldn't figure out how to do it without passing in a param. For this to work it would need access to the params and I didn't know the best way to add arguments to `axios.create`.

I guess I could rewrite the code to accept a param...

Also, I thought the request intercept would happen before the request is made? 🤔

jenbutongit: should've been clearer, sorry, so with intercept what happens is call is made -> axios intercepts it and changes -> request carries on to wherever it's heading 

but ideally it should be request called with everything it needs without the interception

api/services/CasebookService.js
jenbutongit: oops, just read the docs, yeah - it is now legacy and not recommended but it's not deprecated. 

have you had a look at URLSearchParams on the node docs? https://nodejs.org/api/url.html#new-urlsearchparamsobj 

```.js
const params = new URLSearchParams({
  user: 'abc',
  query: ['first', 'second']
});
```

api/services/CasebookService.js
```diff
@@ -1,53 +1,65 @@
...
+const {
+    hmacKey,
+    casebookCertificate: cert,
+    casebookKey: key,
+    customURLs,
+} = config;
 
-    post(options) {
-        return CasebookService._returnRequestMethods('post', options);
+const CasebookService = axios.create({
```
jenbutongit: casebookService should be a collection of methods, not so much the axios request itself

api/services/CasebookService.js
jenbutongit: in the docs/comments for transformRequest
```
  // You may modify the headers object.
  transformRequest: [function (data, headers) {
```

if you can't get it working intercept will be ok

api/controllers/OpenEAppController.js
RichardBray: Good point. For some reason, I thought that casebook would return a 200 but contain an error with a message but I just check the documentation they sent and that would only happen for 400's or 422's. I'll get that sorted.

api/controllers/DashboardController.js
RichardBray: Sure makes sense. I'll try to do it for piping/streaming files for download but it might be a bit difficult.

api/services/CasebookService.js
RichardBray: Yeah, I saw that but annoyingly it doesn't quite do what Casebook wants. The code snippet you posted returns this:
```
'user=abc&query=first%2Csecond'
```

But this is what Casebook wants.
```
user=abc&query=first&query=second'
```

Not sure if `URLSearchParams` has an option for this but I wasn't able to find one.

api/services/CasebookService.js
RichardBray: I've tried it without intercept but by passing an argument in. Let me know your thoughts on that approach.

api/services/CasebookService.js
jenbutongit: the documentation says this:
> Arrays are stringified using array.toString(), which simply joins all array elements with commas.

so what's happening is it's actually trying to encode `{ query: 'first,second' }` 

```.js
// Using an array
const params = new URLSearchParams([
  [ 'user',  'abc' ]
  ['query', 'first'],
  ['query', 'second'],
]);
console.log(params.toString());
// Prints 'user=abc&query=first&query=second'
```
I suppose if you'd prefer to pass in an object rather than tuples, your method is all good 👍

api/controllers/DashboardController.js
```diff
@@ -144,48 +144,30 @@ const dashboardController = {
...
-        };
+    async _getDataFromCasebook(results) {
+        try {
+            const applicationReferences = results.map(
+                (resultItem) => resultItem.unique_app_id
+            );
+            const queryParamsObj = {
+                timestamp: Date.now().toString(),
+                applicationReference: applicationReferences,
+            };
```
jenbutongit: this looks duplicated in `FileDownloadController` as well. 

it seems like to get an application status, you always have to pass a timestamp, as well as application reference(s). `CasebookService.getApplicationStatus` can handle adding timestamp itself. 

```
//(just shorthand, don't add it as an anon function)
CasebookService.getApplicationStatus = (references) => ({

const query = { 
  timestamp: Date.now(),
  applicationReference: references
}
 //..
})
```

api/controllers/DashboardController.js
```diff
@@ -144,48 +144,30 @@ const dashboardController = {
...
-            .catch((err) => {
-                sails.log.error(
-                    `Error returned from Casebook API call: ${err}`
-                );
-            });
+            return await CasebookService.getApplicationStatus(queryParamsObj);
+        } catch (error) {
+            sails.log.error(error);
+            return res.serverError();
+        }
```
jenbutongit: ah, so tidy ❤️

api/controllers/FileDownloadController.js
```diff
@@ -92,41 +88,26 @@ const FileDownloadController = {
...
-            });
-    },
+            sails.log.info('Downloading file from Casebook');
 
-    _renamePDFFromHeader(req, res) {
-        res.headers[
-            'content-disposition'
-        ] = `attachment; filename=LegalisedDocument-${req.params.apostilleRef}.pdf`;
+            const response = await CasebookService.getApostilleDownload(
+                queryParamsObj
```
jenbutongit: same comment as here https://github.com/UKForeignOffice/loi-application-service/pull/305/files#r747725006. `FileDownloadController` doesn't need to concern itself about how casebook api works, it just wants to get the stream for the provided reference

api/controllers/OpenEAppController.js
```diff
@@ -63,67 +61,34 @@ const OpenEAppController = {
...
-                sails.log.error(err);
-                res.serverError();
-            });
+    async _getApplicationDataFromCasebook(req, res) {
+        try {
+            const queryParamsObj = {
+                timestamp: new Date().getTime().toString(),
+                applicationReference: req.params.unique_app_id,
+            };
+            return await CasebookService.getApplicationStatus(queryParamsObj);
```
jenbutongit: https://github.com/UKForeignOffice/loi-application-service/pull/305/files#r747725006

api/controllers/DashboardController.js
RichardBray: Good point. I'm sure there was a time where these objects were different, but maybe I refactored that away 🤷

api/services/CasebookService.js
```diff
@@ -1,52 +1,93 @@
...
+                return response;
+            },
+            (error) => {
+                return Promise.reject(error);
+            }
+        );
 
-        const queryStr = apiQueryString.stringify(queryParamsObj);
+        return casebookRequestBase;
+    },
```
jenbutongit: it was ok as it was 😄!! you shouldn't have to create a new instance each time you want to do a call.

```.js

const baseRequest = axios.create(..) 
baseRequest.interceptors.response.use(...)

const CasebookService = { 
  getApplicationStatus(applicationReference) {
    const query = {
      timestamp: Date.now().toString(),
      applicationReference,
    };
    const params = thatFunctionThatParsesIt(query)
    //.. 
    return casebookRequestBase.get(customURLs.applicationStatusAPIURL, {
             params,
             // if there are other axios options you want to override from the baseRequest, you can do so here
    });
  }
}
```

the idea is that axios creates an instance with the options and makes them the default *for that instance*, so you don't have to create a fresh instance every time yourself.
`baseRequest.post(uri, {...someOtherOptions})` shouldn't change baseRequest and update with the new options.

api/services/CasebookService.js
```diff
@@ -1,52 +1,93 @@
...
-                cert,
-                key,
-            },
-            headers: {
-                hash,
-            },
+        return hash;
+    },
+
+    _addHmacToQueryParam(data, _headers) {
```
jenbutongit: you don't have to change this now because it would affect the consistency of how things are done in the rest of the app, but you don't have to use `_` or `private` fields when you're writing modules this way. imo it only really matters for (class) OOP 

`module.exports = CasebookService` <- so you're only exposing everything within the casebook object and anything you don't want to expose, you can just take it out of the object. 

The alternative is (although you probably don't have the right transpilation/node ver for this), to export the functions you want to expose. 

```.js
const baseRequest = axios(..) // not exported, private 
export async function getApplicationReference { } // exported, public
```
then, for anything that needs casebook service
```.js
const { getApplicationReference } = require('../CasebookService')
// or import the whole thing if you use multiple methods
```

api/services/EmailService.js
```diff
@@ -39,16 +37,11 @@ const EmailService = {
...
-            { url, body: postData },
-            (err, res, body) => {
-                if (err) {
-                    sails.log.error(err);
-                } else {
-                    sails.log.info(res.statusCode, body);
-                }
-            }
-        );
+        emailRequest.post(url,{ data: postData }).then((response) => {
```
jenbutongit, line 40: is url always the same?

api/services/LocationService.js
```diff
@@ -44,17 +46,13 @@ var LocationService = {
     },
 
     postcodeLookup: function postcodeLookup(postcode) {
-      var rp = require('request-promise');
       var options = JSON.parse(JSON.stringify(sails.config.customURLs.postcodeLookUpApiOptions));
```
jenbutongit: probably some old code? This is the old way to clone an object (nowadays you would just do `const = { ...myObj }`. But I can't see why it needs to be cloned tbh.. so suggestion👇 cos it's not your problem (but it's bugging me too much I want to fix it lol) 

```suggestion
     const { uri, timeout } = sails.config.customURLs.postcodeLookUpApiOptions;
     return axios.get(`${ uri }${ postcode }`, { timeout })
```

api/helpers/queryParamObjToStr.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * URLSearchParams does not support converting arrays to query params.
```
jenbutongit:
```suggestion
 * URLSearchParams does not support converting arrays to separate query params.
```
It does support converting arrays to query params, just not how your app requires you to 🙂

api/helpers/queryParamObjToStr.js
```diff
@@ -0,0 +1,23 @@
...
+ * i.e. dooing {test: ['a', 'b']} to '?test=a&test=b'. This solves that.
```
jenbutongit: >dooing

api/services/CasebookService.js
```diff
@@ -1,52 +1,93 @@
...
+            paramsSerializer: queryParamObjToStr,
+            transformRequest: [CasebookService._addHmacToQueryParam],
+        });
+
+        casebookRequestBase.interceptors.response.use(
+            (response) => {
+                if (response.status !== 200) {
+                    throw new Error(`Casebook returned ${response.statusCode}`);
+                }
+                return response;
```
jenbutongit: this means that any other status code, even if it's 2xx (so for example, 204) will result in an error. Is that right? (not a rhetorical question or anything, it might be that even if it does receive a 2xx request, it's still an error for eia)

this is how axios can resolve ranges 
```
axios.get('/user/12345', {
  validateStatus: function (status) {
    return status < 500; // Resolve only if the status code is less than 500
  }
})
```

api/controllers/DashboardController.js
```diff
@@ -144,48 +144,28 @@ const dashboardController = {
...
-                }
-                return response;
-            })
-            .catch((err) => {
-                sails.log.error(
-                    `Error returned from Casebook API call: ${err}`
-                );
-            });
+            return await CasebookService.getApplicationStatus(
+                applicationReferences
```
jenbutongit: nice! I would probably be lazy and just do the .map inside the function (but you definitely should not follow that example cos it tends to get quite unreadable 😄)

jenbutongit (changes requested):

api/controllers/DashboardController.js
jenbutongit: 🪄

api/services/CasebookService.js
jenbutongit: (sorry!! I realise this might actually be a bit of a trial by fire since the request and fetch libaries don't have this concept)

RichardBray:
In response to this question (https://github.com/UKForeignOffice/loi-application-service/pull/305#discussion_r747488476) `transformRequest ` only applies to 'PUT', 'POST', 'PATCH' and 'DELETE' and no 'GET' 😢  so this won't quite work.

api/services/CasebookService.js
RichardBray: In response to this question [(#305 (comment)) ](https://github.com/UKForeignOffice/loi-application-service/pull/305#discussion_r747488476) transformRequest only applies to 'PUT', 'POST', 'PATCH' and 'DELETE' and no 'GET' 😢 so this won't quite work.

api/services/CasebookService.js
RichardBray: Ah, this is definitely currently possible. I think you can do something like:
```js
module.exports.getApplicationStatus = CasebookService.getApplicationStatus;
```
or
```js
module.exports = {
  getApplicationStatus: CasebookService.getApplicationStatus
};
```

I'll try it out.

api/services/CasebookService.js
RichardBray: Ah I didn't know this. I might do this instead actually.

RichardBray: Nah, it's more work than it's worth. I need the object to calculate the hash. I think my function is more appropriate in this specific scenario but this is useful to know for the future.

api/services/CasebookService.js
jenbutongit: if you decide to do `module.exports`, there's no point in having the CasebookService object, because you're already selectively picking which methods you're exposing 

```
module.exports = {
  getApplicationStatus
 // not getApplicationStatus: CasebookService.getApplicationService
};
```
Essentially, there's no "benefit" of wrapping everything in CasebookService. The module is already `casebookService.js` (i.e. require('casebookService'))

api/services/EmailService.js
RichardBray: Nope, it will either be '/confirm-submission' or '/failed-documents' based on the function that is called.

api/services/CasebookService.js
RichardBray: Nah this is for sure an issue. I like the look of `validateStatus` but the default function seems to be exactly what I need:
```
  validateStatus: function (status) {
    return status >= 200 && status < 300; // default
  },
```

Maybe I should just get rid of my implementation 🤷

jenbutongit: the less code you have to write, the better ;)

api/controllers/DashboardController.js
RichardBray: Nah I'm sure there's a way to do it and have readable code.

api/helpers/queryParamObjToStr.js
RichardBray: 🤦

api/services/CasebookService.js
RichardBray: No worries, it's all part of the learning experience 😁

jenbutongit (approved):
LGTM ❤️

PR #309, opened 2021-11-17
Feat/eia 245 character counter

RichardBray:
# Description
Adds character counter and error message to eApp user reference screen.
https://jira.digital.csd.fcdo.gov.uk/browse/EIA-245

Character counter when JS is off:

Screenshot 2021-11-17 at 15 21 02

Character counter when JS is on:
Screenshot 2021-11-17 at 16 47 45
Screenshot 2021-11-17 at 15 20 39

Error screen if user enters more than 30 chars and hits CONTINUE:
Screenshot 2021-11-17 at 16 54 49

I know that the govuk-frontend js files provide a character counter already (my code was based on that), but their docs recommend you add all their JS code which I thought was unecessary. Also, their character counter only supports textareas so I thought it would be better if I put my own together.

## Type of change

Please delete options that are not relevant.

# How Has This Been Tested?

- [x] Manually
- [x] Unit tests

# Checklist:

### DIFF
--- a/assets/js/characterCounter.js
+++ b/assets/js/characterCounter.js
@@ -10,19 +10,25 @@ if (parentElem) {
         '.govuk-character-count__message'
     );
 
-    inputHint.textContent = `You have ${maxlength} characters remaining`;
+    if (inputElem.value.length > 0) {
+        updateRemainingCharacters(maxlength, inputElem, inputHint);
+    } else {
+        inputHint.textContent = `You have ${maxlength} characters remaining`;
+    }
 
-    inputElem.addEventListener('input', (inputEvent) =>
-        updateRemainingCharacters(maxlength, inputEvent, inputHint)
+    inputElem.addEventListener('input', (_e) =>
+        updateRemainingCharacters(maxlength, inputElem, inputHint)
     );
 }
 
-function updateRemainingCharacters(maxlength, inputEvent, inputHint) {
-    const charactersRemaining = maxlength - inputEvent.target.value.length;
+function updateRemainingCharacters(maxlength, inputElem, inputHint) {
+    const charactersRemaining = maxlength - inputElem.value.length;
 
     if (charactersRemaining < 0) {
         inputHint.classList.replace('govuk-hint', 'govuk-error-message');
-        inputHint.textContent = `You have ${Math.abs(charactersRemaining)} characters too many`;
+        inputHint.textContent = `You have ${Math.abs(
+            charactersRemaining
+        )} characters too many`;
     } else {
         inputHint.classList.contains('govuk-error-message') &&
             inputHint.classList.replace('govuk-error-message', 'govuk-hint');
--- a/assets/js/characterCounter.js
+++ b/assets/js/characterCounter.js
@@ -1,37 +1,55 @@
-
-const parentElem = document.querySelector(
+const inputParentElem = document.querySelector(
     '[data-module="govuk-character-count"]'
 );
 
-if (parentElem) {
-    const inputElem = parentElem.querySelector('.govuk-js-character-count');
-    const { maxlength } = parentElem.dataset;
-    const inputHint = parentElem.querySelector(
-        '.govuk-character-count__message'
+if (inputParentElem) {
+    prepareCharCounter();
+}
+
+function prepareCharCounter() {
+    const inputElem = inputParentElem.querySelector(
+        '.govuk-js-character-count'
     );
+    const commonParams = {
+        charLimit: Number(inputParentElem.dataset.maxlength),
+        inputElem,
+        inputHint: inputParentElem.querySelector(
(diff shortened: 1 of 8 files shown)


views/eApostilles/additionalReference.ejs
```diff
@@ -5,22 +5,42 @@
...
+    <%if (inputError) { %>
+        <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1"
+            data-module="govuk-error-summary">
+            <h2 class="govuk-error-summary__title" id="error-summary-title">
+                There is a problem
+            </h2>
+            <div class="govuk-error-summary__body">
+                <ul class="govuk-list govuk-error-summary__list">
+                    <li>
+                        <a href="#user-reference">Enter a reference no longer than 30 characters</a>
```
jenbutongit: The suggested error message from the design system is

>Job description must be 350 characters or fewer

https://design-system.service.gov.uk/components/character-count/error/index.html

unless a content designer or research has suggested otherwise, ideally you would use what the design system have suggested. At the bottom of each of the components, usually there's some text about what the error messages should look like. Usually it's an amalgamation of things that have already been tested, researched, refined etc

assets/js/characterCounter.js
```diff
@@ -0,0 +1,55 @@
...
+    );
+}
+
+function displayInitialHintText(commonParams) {
+    const { charLimit, inputElem, inputHint } = commonParams;
+    const inputHasValue = inputElem.value.length > 0;
+
+    inputHasValue
+        ? updateHintText(commonParams)
+        : (inputHint.textContent = `You have ${charLimit} characters remaining`);
```
jenbutongit: you will need to check if there are polyfills on this project or some sort of transpilation (like babel) on these js files. 

I'm pretty sure template strings and destructure isn't IE11 compatible.. and yes, unfortunately we will have to support IE11.. it's refreshed every 6 months.. so hopefully we don't have to support it for much longer. Worth bookmarking this link! https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices

assets/js/characterCounter.js
```diff
@@ -0,0 +1,55 @@
+const inputParentElem = document.querySelector(
```
jenbutongit: regards to this: 
>I know that the govuk-frontend js files provide a character counter already (my code was based on that), but their docs recommend you add all their JS code which I thought was unecessary. Also, their character counter only supports textareas so I thought it would be better if I put my own together.

I had a look at this https://github.com/alphagov/govuk-frontend/blob/e0d5aad300e84e66b42bffa98fa310d630301099/src/govuk/components/character-count/character-count.js

there's nothing to suggest it won't work for text inputs - it's just using the data-module as a selector. If it doesn't work, you can always just change the selector. 

It seems like the project is already including govuk scripts, so really it's no harm in adding another. If you prefer not to, you can just copy + paste this individual script. 

I do think your code is neater (and modern 💪), but unfortunately they do have a more "complete" solution as well as well tested (they are listening for things like `onBlur` as well as handling the plural for word/s or character/s correctly).

jenbutongit (changes requested):

views/eApostilles/additionalReference.ejs
RichardBray: Ah, I don't know why I didn't notice that before. I pretty much copied the error from the existing service and Liz (our content writer) said it was okay. I will change it and let her know. I'm sure she'll be fine with it.

assets/js/characterCounter.js
RichardBray: Hmm, I guess since it's already installed it wouldn'\t hurt to use the govuk-frontend file. Although I'm not a huge fan of installing all the js like in the readme:
https://frontend.design-system.service.gov.uk/get-started/#4-get-the-font-and-images-working

So I'll try just using the character counter.

jenbutongit: you could just init the one file - although it seems like they do include some other helpers for accessibility. If you really don't like it you can just copy and paste the one file or snippet

RichardBray: Makes sense. I've actually just gone ahead and done that. There are so many polyfill needed for old IE 😢. I remember using the classList one back in the day. Can't believe this browser is still being used.

jenbutongit: "Still being used" is generous, I believe it's 0.7% global usage at the moment so not much longer

RichardBray: really! that small! Why is it still on this list 😭?
https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices#browsers-to-test-in

jenbutongit: global usage is low - but they do base it off of usage of the gov.uk website, which might have a higher %

jenbutongit (approved):

PR #312, opened 2021-11-22
Fix/EIA-267 application download

RichardBray:


api/controllers/FileDownloadController.js
```diff
@@ -46,12 +46,9 @@ const FileDownloadController = {
...
     _apostilleRefBelongToApplication(req, res) {
         FileDownloadController._urlErrorChecks(req, res);
-        const queryParamsObj = {
-            timestamp: Date.now().toString(),
-            applicationReference: req.params.unique_app_id,
-        };
+        const applicationRef = req.params.unique_app_id;
 
-        return CasebookService.getApplicationStatus(queryParamsObj)
+        return CasebookService.getApplicationStatus(applicationRef)
```
jenbutongit, line 51: this wouldn't happen with typescript ;) 

If it continues to be an issue, you can change CasebookService.getApplicationStatus to throw an error if applicationRef is not typeof string. (or is typeof object) 

then you can write test for casebook service, that checks that if you pass in an obj/string/number to see if it throws

jenbutongit (approved):

api/controllers/FileDownloadController.js
RichardBray: Ha, I wish I was using Typescript. Hopefully, this won't be an issue anymore but yeah it makes sense to add some sort of type checking if it does.

PR #315, opened 2021-11-23
feat: add ability to download receipt

RichardBray:
# Description

This ticket add the download receipt button link which makes a GET call to Casebook and streams the PDF file to the client.

Screenshot 2021-11-24 at 15 45 02

## Type of change

# How Has This Been Tested?

- [x] Unit tests
- [x] Manual testing

# Checklist:

- [x] Any dependent changes have been merged and published in downstream modules


views/eApostilles/openEApp.ejs
```diff
@@ -148,4 +148,5 @@
 
         <p>Your payment will appear on your bank statement as FCDO Legalisation Service.</p>
 
-    </section>
\ No newline at end of file
+        <a href="/download-receipt/<%= applicationId %>" class="govuk-link govuk-!-margin-top-2" role="button" download>View and download your receipt</a>
```
jenbutongit: I don't think its necessary to add another layer of abstraction `ReceiptDownloadController` if you're only able to download from one place (for now). I would probably put the ReceiptDownloadController in the openEApp controller (whatever is used to render this page) - because they're all related

jenbutongit (changes requested):

views/eApostilles/openEApp.ejs
RichardBray: Hmm, I was thinking that before but I wanted to make the files more single responsibility if that makes sense. So putting it in openEApp would mean that file is doing multiple things, maybe it could be a service instead of a controller 🤷

jenbutongit: if it's a service it would go into CB service, which it already is

as far as single responsibility goes, yes, moving it to openEAppController would be correct. From what I can tell the app is using the MVC pattern. really simplistically.. Model would the casebook api, view would be the templates, and controller would broker interactions between the view and model. 

openEAppController wouldn't be single responsibility if it was rendering another page for example, or handling a stream that was completely unrelated.

files don't have to be single responsibility themselves

RichardBray: Fair enough I see where you're coming from. I've updated the code accordingly. Let me know your thoguhts.

api/controllers/OpenEAppController.js
```diff
@@ -150,6 +151,42 @@ const OpenEAppController = {
...
+            throw new Error(`downloadReceipt Error: ${err}`);
+        }
+    },
+
+    _errorChecks(req, res) {
+        const userData = HelperService.getUserData(req, res);
+        if (!userData.loggedIn) {
+            throw new Error('User is not logged in');
+        }
+        if (req.params.applicationRef === 'undefined') {
```
jenbutongit, line 196: is applicationRef === 'undefined'? or typeof applicationRef === 'undefined'?

api/controllers/OpenEAppController.js
```diff
@@ -150,6 +151,42 @@ const OpenEAppController = {
...
+    async downloadReceipt(req, res) {
+        try {
+            await OpenEAppController._streamReceiptToClient(req, res);
+        } catch (err) {
+            sails.log.error(err);
+            return res.serverError();
+        }
+    },
+
+    async _streamReceiptToClient(req, res) {
```
jenbutongit, line 169: do you have a few streams you have to handle? (you can leave it for now - but might be worth considering a helper function that handles streams generically)

tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -321,4 +323,48 @@ describe('OpenEAppController', () => {
...
+        it('triggers serverError if application ref is undefined', async () => {
+            // when
+            sandbox.stub(HelperService, 'getUserData').callsFake(() => ({
+                loggedIn: true,
+            }));
+            reqStub.params.applicationRef = undefined;
+            await OpenEAppController.downloadReceipt(reqStub, resStub);
+
+            // then
+            expect(resStub.serverError.calledOnce).to.be.true;
```
jenbutongit: is the error message `'Missing application reference'`?  

the error as it is, should only throw if `reqStub.params.applicationRef = 'undefined' //string containing the word undefined` not `undefined` (the keyword).

jenbutongit (changes requested):

api/controllers/OpenEAppController.js
RichardBray: this is the second one, AFAIK there shouldn't be anymore but you never know 🤷

api/controllers/OpenEAppController.js
RichardBray: This is down to the wierness of ejs. SO this is part of the code that triggers the route:
```ejs
<a href="/download-receipt/<%= applicationId %>"
```
If `applicationId` is unefined it will return it in this url string so you get the following:
Screenshot 2021-11-25 at 18 12 37

And so it ends up as a string in the js code:
Screenshot 2021-11-25 at 18 12 48

api/controllers/OpenEAppController.js
RichardBray: Good point. I will make that change 👍

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Good point. I will make that change 👍

RichardBray: Yes, the error message is `'Missing application reference'`. I did originally think about testing the throw but I've only exposed one function and the other unexposed function is the one that actually does the throwing. So I'm not sure how to go about testing that without exposing it as well.

api/controllers/OpenEAppController.js
jenbutongit: ah I see, what's happening is something like this `download-receipt/${ref}` - ref is undefined, but if you wrap it in a string, it will render `download-receipt/undefined`. 

a couple things.. 

1. what are the cases that will cause applicationId to be undefined
  - also, would it be worth adding an if statement so people aren't greeted with an error page? (only render when applicationId is defined) 
2. I'm assuming this is all auth'd? what is the risk that I navigate to `/download-receipt/random-ref` and download someone else's receipt? is there any risk in that? (am I downloading someone else's data?)

jenbutongit (changes requested):

tests/specs/controllers/OpenEAppController.test.js
jenbutongit: that's ok if you're not "close" enough to a function that throws. I just wanted to make sure that test was throwing for the reason you were expecting (if that made any sense)

RichardBray:
>ah I see, what's happening is something like this download-receipt/${ref} - ref is undefined, but if you wrap it in a string, it will render download-receipt/undefined.

Yep, exactly

> 1. what are the cases that will cause applicationId to be undefined

This should only really happen if someone messes around with the download url and gets rid of it.

> 2. I'm assuming this is all auth'd? what is the risk that I navigate to /download-receipt/random-ref and download someone else's receipt? is there any risk in that? (am I downloading someone else's data?)

Good point. Shouldn't be too difficult to add.

jenbutongit (approved):
🎉

PR #317, opened 2021-11-29
Feat/eia 286 hide info rejected

RichardBray:
## Description

This ticket essentially hides all the important copy from an application page if all the documents are rejected because we currently don't have any text for it.

Screenshot 2021-11-29 at 11 30 34

---

Instead of this:

Screenshot 2021-11-30 at 12 07 24

---

This is supposed to be a quick and dirty fix, at some point we will get a content p[erson to write some copy specifically for this scenario.

### DIFF
--- a/api/controllers/OpenEAppController.js
+++ b/api/controllers/OpenEAppController.js
@@ -33,30 +33,33 @@ const OpenEAppController = {
                     req,
                     res
                 );
+            const [casebookData] = casebookResponse;
+            const casebookStatus = casebookData.status || 'Not available';
+            const casebookDocuments = casebookData.documents || [];
+
             const pageData = OpenEAppController._formatDataForPage(
                 applicationTableData,
-                casebookResponse[0]
+                casebookData
             );
             const userRef = await OpenEAppController._getUserRef(
-                casebookResponse[0],
+                casebookData,
                 res
             );
             const daysLeftToDownload =
-                casebookResponse[0].status === 'Completed'
+                casebookData.status === 'Completed'
                     ? OpenEAppController._calculateDaysLeftToDownload(
-                          casebookResponse[0],
+                          casebookData,
                           req
                       )
                     : 0;
             const applicationExpired =
                 OpenEAppController._hasApplicationExpired(
-                    casebookResponse[0],
+                    casebookData,
                     daysLeftToDownload
                 );
 
-            const noOfRejectedDocs = OpenEAppController._calculateRejectedDocs(
-                casebookResponse[0]
-            );
+            const noOfRejectedDocs =
+                OpenEAppController._calculateRejectedDocs(casebookData);
 
             res.view('eApostilles/openEApp.ejs', {
                 ...pageData,
@@ -64,9 +67,9 @@ const OpenEAppController = {
                 user_data: userData,
                 daysLeftToDownload,
                 applicationExpired,
-                applicationStatus: casebookResponse[0].status,
+                applicationStatus: casebookStatus,
                 allDocumentsRejected:
-                    noOfRejectedDocs == casebookResponse[0].documents.length,
+                    noOfRejectedDocs == casebookDocuments.length,
             });
         } catch (error) {
             sails.log.error(error);
@@ -90,6 +93,13 @@ const OpenEAppController = {
         if (!casebookResponse) {
             throw new Error('No data received from Casebook');
         }
+        if (!casebookResponse.documents) {
+            throw new Error('No documents found from Casebook');
+        }
+        if (!casebookResponse.payment) {
+            throw new Error('No payment info found from Casebook');
+        }
+
         return {
             applicationId: applicationTableData.unique_app_id,
             dateSubmitted: OpenEAppController._formatDate(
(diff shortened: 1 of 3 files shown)


api/controllers/OpenEAppController.js
```diff
@@ -55,13 +54,19 @@ const OpenEAppController = {
...
+
             res.view('eApostilles/openEApp.ejs', {
                 ...pageData,
                 userRef,
                 user_data: userData,
                 daysLeftToDownload,
                 applicationExpired,
                 applicationStatus: casebookResponse[0].status,
+                allDocumentsRejected:
+                    noOfRejectedDocs == casebookResponse[0].documents.length,
```
alisalaman-cyb: probably needs a guard condition to protect against casebookResponse[0] = null / undefined OR casebookResponse[0].documents = null / undefined.  There are also similar cases throughout so my first point would be to check instances where you're retrieving casebookResponse[0] and check if this is possible.  Could be a red herring but definitely worth checking to avoid being bitten.

api/controllers/OpenEAppController.js
```diff
@@ -92,7 +97,7 @@ const OpenEAppController = {
             ),
             documents: casebookResponse.documents,
             originalCost: HelperService.formatToUKCurrency(
-                casebookResponse.payment.netAmount
+                casebookResponse.payment.transactions[0].amount
```
alisalaman-cyb: could casebookResponse.payment.transactions be unset here?  Would recommend a guard condition needed here with a fallback value - 0?  Also formatToUKCurrency probably covers it but worth checking if it always include two decimal places.

api/controllers/OpenEAppController.js
```diff
@@ -202,6 +206,16 @@ const OpenEAppController = {
             throw new Error('User not authorised to download this receipt');
         }
     },
+
+    _calculateRejectedDocs(casebookResponse) {
+        let rejectedDocs = 0;
+        for (const document of casebookResponse.documents) {
```
alisalaman-cyb, line 236: could casebookResponse.documents be undefined here?  Probably needs a guard condition to check this first then return 0 if so

alisalaman-cyb (commented):
Hey @RichardBray , some feedback provided.  I think you probably need to add some guard checks in there to avoid edge cases where values are not defined...

api/controllers/OpenEAppController.js
RichardBray: Yep for sure, luckily `formatToUKCurrency` uses the [internatianlization api](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) to format numbers so that should take care of the two decimal places 🤞

api/controllers/OpenEAppController.js
RichardBray: It's not showing as outdated in GIthub but the code for this section has definitely changed:
https://github.com/UKForeignOffice/loi-application-service/pull/317/files#diff-9e383893e624afb9360415d5317a4a07c9360970f77ec0dc5ba0e9bef6e4c784R225

api/controllers/OpenEAppController.js
```diff
@@ -90,20 +93,31 @@ const OpenEAppController = {
...
+
         return {
             applicationId: applicationTableData.unique_app_id,
             dateSubmitted: OpenEAppController._formatDate(
                 applicationTableData.createdAt
             ),
             documents: casebookResponse.documents,
             originalCost: HelperService.formatToUKCurrency(
-                casebookResponse.payment.transactions[0].amount
+                casebookResponse.payment.transactions[0].amount || 0
```
alisalaman-cyb, line 116: I see you've added the guard condition above for when payment is not specified but you probably need the same for when payment.transactions are not specified as well - not sure if that can even happen but worth checking

api/controllers/OpenEAppController.js
```diff
@@ -33,40 +33,43 @@ const OpenEAppController = {
...
                 ...pageData,
                 userRef,
                 user_data: userData,
                 daysLeftToDownload,
                 applicationExpired,
-                applicationStatus: casebookResponse[0].status,
+                applicationStatus: casebookStatus,
                 allDocumentsRejected:
-                    noOfRejectedDocs == casebookResponse[0].documents.length,
+                    noOfRejectedDocs == casebookDocuments.length,
```
alisalaman-cyb: May want to consider changing these to === to check the type as well - minor but good practice

api/controllers/OpenEAppController.js
alisalaman-cyb: Thanks, I can see the changes reflected in the actual commit: https://github.com/UKForeignOffice/loi-application-service/pull/317/commits/e9408622342e7a9c2df7acacfd78266c64852c40

alisalaman-cyb (commented):
Hey @RichardBray , thanks for addressing those issues.  Minor points returned back for you to check.

api/controllers/OpenEAppController.js
RichardBray: Good spot 👍

api/controllers/OpenEAppController.js
RichardBray: Yeah, it makes sense. Max told me about this library (https://github.com/gcanti/io-ts) which makes sure the response from the API matches a specific object and if it doesn't, it responds with a failure. Unfrountatley this app wasn't written in TypeScript but if it was then this library would be very useful.

alisalaman-cyb (approved):
LGTM

PR #320, opened 2021-12-13
Feat/eia 282 upload spinner

RichardBray:
## Old loader [DEPRECATED]
https://user-images.githubusercontent.com/1377253/145816794-0bbac180-6fc7-4c22-a18c-e71e75349b56.mp4

## Updated loader
https://user-images.githubusercontent.com/1377253/148226176-4251d5d3-e8e1-4771-8dae-8c4976556380.mp4


jenbutongit:
I think this one might be worth having a chat with Alvin or Dan (to see if it's worth changing this any further). I think a progress bar would be better for uploads (or could be a spinner with text), since a user will be able to tell if the upload is hung or is actually progressing. Is this feature meant to circumvent slower uploads?

This one might be a good candidate for user testing. Would taking a user away from the upload page, to the spinner, then back to the upload page confuse them? The page looks exactly the same (apart from the bottom). Is there a risk a user will try to re upload the files (since you're giving them the button?). What happens when you upload again? does it just add more to the list?

RichardBray:
> This one might be a good candidate for user testing. Would taking a user away from the upload page, to the spinner, then back to the upload page confuse them? 

Good point about hiding and showing page again, the user could get confused about what has just happened.

> The page looks exactly the same (apart from the bottom). Is there a risk a user will try to re upload the files (since you're giving them the button?). What happens when you upload again? does it just add more to the list?

Fair point, I guess if we were going down this route we'd have to highlight what has changed. Also if the user does upload the same file twice they should get an error but I get why you ask.

I'll ping the others tomorrow in standup and go back to the drawing board on this one.

views/eApostilles/uploadFiles.ejs
```diff
@@ -80,24 +80,31 @@
...
                 </button>
             </form>
         </div>
 
+        <div class="js-progress-bar govuk-!-display-none govuk-!-margin-bottom-4" aria-live="polite">
+            <label class="js-upload-progress-text form-label-bold" for="upload-progress-bar">
+                Uploading...
+            </label>
+            <div class="upload-progress--outline"></div>
+            <div class="js-upload-progress-bar upload-progress--bar" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100"></div>
```
jenbutongit: are the aria-values actually being updated? you may need to throttle your network speed to see if it's getting updated properly. they need to be updated, since screen readers will rely on this for the progress

assets/js/uploadFileProgressBar.js
```diff
@@ -0,0 +1,143 @@
...
+    var trident = ua.indexOf('Trident/');
+
+    return msie > 0 || trident > 0;
+}
+
+if (!browserIsIE()) {
+    UploadProgressBar.init();
+
+    /**
+     * Display time remaining
```
jenbutongit: why time, out of interest? The progress bar seems to be reporting overall progress, eg, uploading, scanning etc, but we're only factoring in the upload part

assets/styles/informed.scss
```diff
@@ -2044,3 +2044,35 @@ caption {
...
+
+.upload-progress--outline {
+  border: 1px solid #1d70b8;
+  height: 30px;
+  position: absolute;
+  width: 100%;
+  box-sizing: border-box;
+}
+
+.upload-progress--bar {
```
jenbutongit: for BEM I would probably expect these to be elements. 

block: `.upload-progress` which encapsulates all of these styles
elements:
- `.upload-progress__outline` 
- `.upload-progress__bar`

modifier:
- `.upload-progress__bar--stripes` 

generally if you have modifiers before elements, eg `.upload-progress--bar__stripes` then it's a sign that things need to be split up differently.


also, `.upload-progress__outline` seems to be an empty div. could it be something like this
```
.upload-progress__bar--outline {
  &:before(or after) {
  ...
  }
}
```
and use it like so
```
<div class="js-upload-progress-bar upload-progress__bar upload-progress__bar--outline" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100">
```

jenbutongit (commented):
ooh first PR of the new year 🎉. I've got a couple Qs. looking good so far. Also, what happened to the package lock? did package.json change in a previous pr?

RichardBray:
> ooh first PR of the new year 🎉. I've got a couple Qs. looking good so far. Also, what happened to the package lock? did package.json change in a previous pr?

Good point. I updated a few packages on a different branch and didn't re-run npm install when I checked out to this branch but merged some changes in, for some reason `npm i` on this project that takes ages to run. I think I might have to do it just to clear up this issue.

views/eApostilles/uploadFiles.ejs
RichardBray: Fair question. I added [this line](https://github.com/UKForeignOffice/loi-application-service/pull/320/files#diff-153f3ee03b38debd99cd3bb44a5151c3a4c9e42eb73378c8617c8c480d8f51bbR51) to update it with client-side js, I don't know if that means I should take off the `aria-valuenow="0"` attribute in the div element 🤷.

Annoyingly it looks like uploading blocks the main thread which made it difficult to inspect the element or write code in the devTools console. So I wrote a little console.log:
```js
console.log(document.querySelector('.js-upload-progress-bar').getAttribute('aria-valuenow'), 'valuenow');
```

Which printed out the following:
Screenshot 2022-01-06 at 11 05 42

I didn't test it on a screen reader or anything so I hope it works 🤞

assets/js/uploadFileProgressBar.js
RichardBray: Time is displayed purely because that's what Alvin suggested in his original design:

Screenshot 2022-01-06 at 11 46 19

But you do raise a good point about only showing times for the upload part. It's difficult to predict how long file scanning will take so I didn't factor that in, but it does actually make sense to not display the time remaining 🤔 

I'll raise it with Alvin.

assets/styles/informed.scss
RichardBray: Hmm, I understand the renaming but I don't understand the benefit of the `upload-progress__bar ` class since it doesn't contain any styles. Is it purely for convention?

jenbutongit: not sure I follow. your .upload-progress--bar has styles, but it should be named ` .upload-progress__bar`

RichardBray: Sorry, I read the code wrong I saw:
```html
<div class="js-upload-progress-bar upload-progress__bar upload-progress__bar--outline"
```
as
```html
<div class="js-upload-progress-bar upload-progress upload-progress__bar"
```
🤦 
Still slowly getting into things after the Christmas break :)

assets/js/uploadFileProgressBar.js
RichardBray: We've ended up removing the seconds timer :(

https://user-images.githubusercontent.com/1377253/148399019-769685e5-901c-4a0d-8cec-f3e2006b7895.mp4

assets/styles/informed.scss
RichardBray: Ha, the only times I've ever used the pseudo-classes `:before` or `:after` is when I've copied css from someone and need to edit it, never actually looked into how they work. Thanks for the tip 👍

jenbutongit (approved):

PR #325, opened 2022-01-06
Feat/EIA-280 tidy list table for mobile

RichardBray:
## Before
Screenshot 2022-01-06 at 16 31 33

## After

Screenshot 2022-01-06 at 16 56 20

### DIFF
--- a/views/partials/dashboardResultsEApp.ejs
+++ b/views/partials/dashboardResultsEApp.ejs
@@ -17,7 +17,7 @@
             <th><a class="sortableHeader govuk-link
         <% if (sortOrder == 1) { %> headerSortDown<%}%>  <% if (sortOrder == -1) { %> headerSortUp<%}%>"
                     href="/dashboard?searchText=<%= searchCriteria %>&sortOrder=<% if (sortOrder == 1) { %>-<%}%>1">
-                    Date submitted <span class="visually-hidden">
+                    Submitted <span class="visually-hidden">
                         <% if (sortOrder==1) { %> ascending<%}%>
                                 <% if (sortOrder==-1) { %> descending<%}%>
                     </span>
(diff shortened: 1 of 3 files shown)


views/partials/dashboardResultsEApp.ejs
```diff
@@ -87,13 +76,13 @@
                         </a>
                     <% } %>
                 </td>
-                <td class="appService hide-col-sm hide-col-md">
+                <td class="hide-col-sm hide-col-md">
                     <%= users_applications[i].user_ref || 'Not available' %>
                 </td>
-                <td class="appService hide-col-sm hide-col-md">
```
alisalaman-cyb, line 93: have you considered using the out of the box styles from the gov design system, e.g. "govuk-table__cell govuk-table__cell--numeric" for numeric cells which automatically right aligns the content?  Maybe easier than trying to code your own in informed.scss

https://design-system.service.gov.uk/components/table/

alisalaman-cyb (commented):
Minor point about using the standard govuk design system styles: https://design-system.service.gov.uk/components/table/

views/partials/dashboardResultsEApp.ejs
RichardBray: Hey Ali, this is a good question. Unfortunately, this project is so old it uses an older version of the govuk sass classes which don't support the newer CSS classes for tables.

https://github.com/alphagov/govuk_elements/blob/master/assets/sass/elements/_tables.scss 

I guess we could update it but I don't want to mess up the existing tables. I think I might add a ticket to update it at some point in the future and find out if Kainos are okay to check my code, to make sure I haven't missed anything out. Would be a big job though 😅

RichardBray: Actually, I take that back, it looks like this project has both the newer style code from `govuk_frontend_toolkit` and the old code from `govuk-elements-sass`, but since the older styles are at the bottom they will overwrite the new styles, so the proper way to fix it would be to remove all the old styles. Lol this project is a mess 🤦 

Anyway... it looks like the old styles apply to the `table` HTML element itself, but the new styles apply to classes added to the `table` element, so if add the new styles without removing the old styles I'll just be adding two sets of styles to the table and the old styles will overwrite the new ones. Sorry just thinking out loud.

alisalaman-cyb (approved):
LGTM

PR #327, opened 2022-01-10
Feat/eia 316 upload no files

RichardBray:
What happens when you click on the upload button without selecting any files to upload?

## Before
Screenshot 2022-01-10 at 17 05 05

## After
Screenshot 2022-01-10 at 16 59 13

### DIFF
--- a/views/eApostilles/openEApp.ejs
+++ b/views/eApostilles/openEApp.ejs
@@ -150,5 +150,5 @@
 
         <a href="/download-receipt/<%= applicationId %>" class="govuk-link govuk-!-margin-top-2" role="button" download="<%= applicationId %>.pdf">Download your receipt</a>
 
-        <a href="/test-download" class="govuk-link govuk-!-margin-top-2" role="button" download="test.png">Test download</a>
+        <a href="/test-download" class="govuk-link govuk-!-margin-top-2" role="button">Test download</a>
     </section>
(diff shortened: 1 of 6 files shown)


views/eApostilles/openEApp.ejs
```diff
@@ -150,5 +150,5 @@
 
         <a href="/download-receipt/<%= applicationId %>" class="govuk-link govuk-!-margin-top-2" role="button" download="<%= applicationId %>.pdf">Download your receipt</a>
 
-        <a href="/test-download" class="govuk-link govuk-!-margin-top-2" role="button" download="test.png">Test download</a>
+        <a href="/test-download" class="govuk-link govuk-!-margin-top-2" role="button">Test download</a>
```
jenbutongit, line 153: what's test download for? is it meant to be merged?

jenbutongit (commented):
lgtm - just one question around test download button

views/eApostilles/openEApp.ejs
RichardBray: ah, so this is a completely separate issue not related to this PR. Basically, we have an issue (which you probably know about) where if a request takes longer than 5 seconds it tries again up to 3 times. In order to test this in AWS I would usually create the code on a new branch and promote that branch to be deployed via Teraform. 

However this time round there were loads of changes made to the Terraform files that would have taken too long to try and fix, so I applied the changes directly onto the `demo` branch with the intention of removing it at the end of the sprint (so tomorrow).

The change you're commenting on was one I accidentally made on the wrong branch so it actually shouldn't be in this pr :)

jenbutongit (approved):

PR #333, opened 2022-01-17
Demo to UAT

RichardBray:
# Changes made
- Download payment receipt (https://consular.atlassian.net/browse/EIA-123)
- Replace depreciated moment.js dependency (https://consular.atlassian.net/browse/EIA-151)
- Automated accessibility testing (https://consular.atlassian.net/browse/EIA-208)
- Replace depreciated request package with a different http package (https://consular.atlassian.net/browse/EIA-190)
- Prevent user from adding more than 30 characters in ref input (https://consular.atlassian.net/browse/EIA-239)
- Show users how many characters they have left in the ref input (https://consular.atlassian.net/browse/EIA-245)
- No 'internal ref' field in summary view when none specified (https://consular.atlassian.net/browse/EIA-266)
- Indication that user clicked 'upload' button (https://consular.atlassian.net/browse/EIA-282)
- Tidy up application list table (https://consular.atlassian.net/browse/EIA-280)
- Download payment receipt (https://consular.atlassian.net/browse/EIA-123)

# Video explanations

The second one is quite long so please feel free to go through the video at 2x speed

https://user-images.githubusercontent.com/1377253/150379915-0f9bcde2-73f6-42df-b247-e1f187b9e1a7.mp4

https://user-images.githubusercontent.com/1377253/150379969-362e60a1-1580-4114-9a65-6fce252305ca.mp4


api/helpers/uploadedFileErrorChecks.js
```diff
@@ -30,13 +30,17 @@ async function connectToClamAV(req) {
...
-    const { clamav_host: clamavHost, clamav_port: clamavPort } =
-        req._sails.config.upload;
-
+    const {
+        clamav_host: clamavHost,
+        clamav_port: clamavPort,
+        clamav_debug_enabled: clamavDebugEnabled,
+    } = req._sails.config.upload;
     const clamAvOptions = {
+        debugMode: JSON.parse(clamavDebugEnabled),
```
simonmoore100: You've declared clamav_debug_enabled in the bit above but then don't use it.
In this line you do a JSON.parse on the env variable but in the bit above you don't.

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -30,13 +30,17 @@ async function connectToClamAV(req) {
 }
 
 function initialiseClamScan(req) {
-    const { clamav_host: clamavHost, clamav_port: clamavPort } =
-        req._sails.config.upload;
-
+    const {
+        clamav_host: clamavHost,
+        clamav_port: clamavPort,
+        clamav_debug_enabled: clamavDebugEnabled,
```
simonmoore100, line 36: Can you make it so clamav_debug_enabled is defaulted to false?

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -30,13 +30,17 @@ async function connectToClamAV(req) {
...
+    } = req._sails.config.upload;
     const clamAvOptions = {
+        debugMode: JSON.parse(clamavDebugEnabled),
         clamdscan: {
             host: clamavHost,
             port: clamavPort,
+            active: true,
```
simonmoore100: What does active mean? There's also clamav_enabled, does it do the same thing?

api/helpers/uploadedFileErrorChecks.js
RichardBray: Sorry, this is just me being confusing. On line **36** I'm renaming `clamav_debug_enabled` to `clamavDebugEnabled` purely for preference, the syntax is a bit weird for that. 

Similar to the way `clamav_host` has been renamed to `clamavHost` and used on line **42**.
```js
const { bar: foo } = { bar: 123 };
bar // Uncaught ReferenceError: bar is not defined
foo // 123
```

But I can take it out if you want.

api/helpers/uploadedFileErrorChecks.js
RichardBray: Sure I can do that. Ideally, the `false` value should be coming from the .env file but having a hardcoded `false` could act as a fallback.

api/helpers/uploadedFileErrorChecks.js
RichardBray: This works a bit differently from `clamav_enabled ` but it's actually pretty minor so I can remove it if necessary.

With **debug mode** turned on it becomes obvious that clamscan tries to search for both the clamscan and clamdscan binaries which don't exist because we are using ClamAV via a socket/TCP connection, not locally.

Screenshot 2022-01-25 at 16 44 16

I thought setting active: true would fix this but it should actually be `false` (https://github.com/kylefarris/clamscan#getting-started). 🤦

I'll push that fix soon.

simonmoore100 (approved):

PR #334, opened 2022-01-24
chore: add proper titles to eApp pages...

RichardBray:
fix spelling mistake in first eligibility qustion url

### DIFF
--- a/views/eApostilles/eligibilityQuestionThree.ejs
+++ b/views/eApostilles/eligibilityQuestionThree.ejs
@@ -1,4 +1,4 @@
-<% pageTitle="Have the PDFs been digitally signed by a notary? - Get a document legalised – GOV.UK" %>
+<% pageTitle="Have the PDFs been digitally signed by a notary?" %>
 
     <div class="inner_header <%if (!user_data.loggedIn){%>no-user-signed-in<%}%>">
         <a href="/eligibility/documents-eligible-for-service" class="back-to-previous govuk-link">Back</a>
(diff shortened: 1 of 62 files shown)


views/eApostilles/eligibilityQuestionThree.ejs
```diff
@@ -1,4 +1,4 @@
-<% pageTitle="e-Apostille eligibility - step 3 of 3" %>
+<% pageTitle="Have the PDFs been digitally signed by a notary? - Get a document legalised – GOV.UK" %>
```
RichardBray: H1 title the page title is referring to.
Screenshot 2022-01-24 at 14 24 41

views/eApostilles/eligibilityQuestionTwo.ejs
```diff
@@ -1,7 +1,7 @@
-<% pageTitle="e-Apostille eligibility - step 2 of 3" %>
+<% pageTitle="Are the documents eligible for this service - Get a document legalised – GOV.UK" %>
```
RichardBray: Screenshot 2022-01-24 at 14 24 34

views/eApostilles/eligibilityQuestionThree.ejs
jenbutongit: I think it would be a better idea to put  "Get a document legalised – GOV.UK" in [layout.ejs ](https://github.com/UKForeignOffice/loi-application-service/blob/develop/views/layout.ejs). That way if the service name changes, there's only one place you have to change it. It will also add the service name to the title for all the pages as well

jenbutongit (changes requested):

jenbutongit (approved):

PR #335, opened 2022-01-28
Look into issue where ClamAV aborts mid scan

RichardBray:
# Description
This ticket is here to achieve a few things
- Fix the issue where checking a documents file type then scanning causes the scan to abort if successful
- Better error handling so that if a scan is actually aborted the file is deleted and the user is informed

## First issue
To fix that the [tokenizer/S3](https://github.com/Borewit/tokenizer-s3) package was used to check the file type instead of checking file type from stream as that was causing issues. 
Tokenizer/S3 is using a [newer version of the aws sdk](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/index.html) (with different syntax) so this was updated in this PR as well but only in one file. I will create a follow-up ticket to update this everywhere.

## Second issue
There are two ways to handle errors to the user for this feature. One refers to the current page and displays the error to the user because there is something they can to do fix it, i.e. this file is not a pdf
The other one shows the following page:

Screenshot 2022-02-01 at 15 03 14

Because the user can't usually do anything like ClamAV is down or aborts the scan.

---

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -4,10 +4,11 @@ const sails = require('sails');
 const uploadFileToStorage = require('../helpers/uploadFileToStorage');
 const deleteFileFromStorage = require('../helpers/deleteFileFromStorage');
 const {
-    virusScanAndCheckFiletype,
+    virusScan,
     checkTypeSizeAndDuplication,
     displayErrorAndRemoveLargeFiles,
     connectToClamAV,
+    checkFileType,
 } = require('../helpers/uploadedFileErrorChecks');
 
 const FORM_INPUT_NAME = 'documents';
@@ -74,7 +75,9 @@ const FileUploadController = {
             }
             sails.log.error(err);
         } else {
-            await virusScanAndCheckFiletype(req);
+            await checkFileType(req);
+            await virusScan(req);
+
             !inDevEnvironment &&
                 FileUploadController._addS3LocationToSession(req);
         }
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -75,14 +75,25 @@ const FileUploadController = {
             }
             sails.log.error(err);
         } else {
+            await FileUploadController._fileTypeAndVirusScan(req, res);
+        }
+
+        FileUploadController._redirectToUploadPage(res);
+    },
+
+    async _fileTypeAndVirusScan(req, res) {
+        try {
             await checkFileType(req, res);
             await virusScan(req, res);
 
             !inDevEnvironment &&
                 FileUploadController._addS3LocationToSession(req);
+        } catch (err) {
+            sails.log.error(err);
+            if (!err.message.includes('STAYONPAGE')) {
+                res.view('eApostilles/fileUploadError.ejs');
+            }
         }
-
-        FileUploadController._redirectToUploadPage(res);
     },
 
     _addS3LocationToSession(req) {
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -75,8 +75,8 @@ const FileUploadController = {
             }
             sails.log.error(err);
         } else {
-            await checkFileType(req);
-            await virusScan(req);
+            await checkFileType(req, res);
+            await virusScan(req, res);
 
             !inDevEnvironment &&
                 FileUploadController._addS3LocationToSession(req);
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -40,11 +40,9 @@ const FileUploadController = {
     uploadFileHandler(req, res) {
         FileUploadController._clearExistingErrorMessages(req);
         const uploadFileWithMulter = FileUploadController._multerSetup(req);
-        uploadFileWithMulter(req, res, async (err) => {
+        uploadFileWithMulter(req, res, (err) => {
             sails.log.info('File successfully uploaded.');
             FileUploadController._errorChecksAfterUpload(req, res, err);
-            await FileUploadController._fileTypeAndVirusScan(req, res);
-            FileUploadController._redirectToUploadPage(res);
         });
     },
 
@@ -69,7 +67,7 @@ const FileUploadController = {
         req.session.eApp.uploadMessages.noFileUploadedError = false;
     },
 
-    _errorChecksAfterUpload(req, res, err) {
+    async _errorChecksAfterUpload(req, res, err) {
         try {
             if (req.files.length === 0) {
                 req.session.eApp.uploadMessages.noFileUploadedError = true;
@@ -81,6 +79,7 @@ const FileUploadController = {
         } catch (err) {
             sails.log.error(err);
             FileUploadController._redirectToUploadPage(res);
+            return;
         }
 
         if (err) {
@@ -92,6 +91,9 @@ const FileUploadController = {
                 sails.log.error(err);
                 res.serverError(err);
             }
+        } else {
+            await FileUploadController._fileTypeAndVirusScan(req, res);
+            FileUploadController._redirectToUploadPage(res);
         }
     },
 
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -1,3 +1,4 @@
+// @ts-check
 const multer = require('multer');
 const sails = require('sails');
 
@@ -6,9 +7,10 @@ const deleteFileFromStorage = require('../helpers/deleteFileFromStorage');
 const {
     virusScan,
     checkTypeSizeAndDuplication,
-    displayErrorAndRemoveLargeFiles,
+    removeLargeFiles,
     connectToClamAV,
     checkFileType,
+    UserAdressableError
 } = require('../helpers/uploadedFileErrorChecks');
 
 const FORM_INPUT_NAME = 'documents';
@@ -74,22 +76,22 @@ const FileUploadController = {
                 throw new Error('No files were uploaded.');
             }
 
-            displayErrorAndRemoveLargeFiles(req);
-
-            if (err) {
-                const fileLimitExceeded =
-                    err.code === MULTER_FILE_COUNT_ERR_CODE;
-                if (fileLimitExceeded) {
-                    req.session.eApp.uploadMessages.fileCountError = true;
-                } else {
-                    res.serverError(err);
-                }
-                throw new Error(err);
-            }
+            removeLargeFiles(req);
 
         } catch (err) {
             sails.log.error(err);
             FileUploadController._redirectToUploadPage(res);
+        }
+
+        if (err) {
+            const fileLimitExceeded =
+                err.code === MULTER_FILE_COUNT_ERR_CODE;
+            if (fileLimitExceeded) {
+                req.session.eApp.uploadMessages.fileCountError = true;
+            } else {
+                sails.log.error(err);
+                res.serverError(err);
+            }
         }
     },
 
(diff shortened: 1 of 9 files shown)


api/controllers/FileUploadController.js
```diff
@@ -61,25 +64,50 @@ const FileUploadController = {
...
+            if (err) {
+                const fileLimitExceeded =
+                    err.code === MULTER_FILE_COUNT_ERR_CODE;
+                if (fileLimitExceeded) {
+                    req.session.eApp.uploadMessages.fileCountError = true;
+                } else {
+                    res.serverError(err);
+                }
+                throw new Error(err);
             }
```
jenbutongit: should this be outside of the try block? `err` is passed in as a parameter. it doesn't rely on anything inside the try block

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -37,49 +44,95 @@ function initialiseClamScan(req) {
...
             host: clamavHost,
             port: clamavPort,
         },
     };
-
-    return new NodeClam().init(clamAvOptions);
+    try {
+        return new NodeClam().init(clamAvOptions);
+    } catch (err) {
+        throw new Error();
```
jenbutongit: can we just throw the `err` instead of throwing a new error with no message?

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -37,49 +44,95 @@ function initialiseClamScan(req) {
...
+async function checkFileType(req) {
     try {
-        if (req.files.length === 0) {
-            req.session.eApp.uploadMessages.noFileUploadedError = true;
-            throw new Error('No files were uploaded.');
+        sails.log.info('Checking file type...');
+        for (const file of req.files) {
+            inDevEnvironment
+                ? await checkLocalFileType(file, req)
+                : await checkS3FileType(file, req);
```
jenbutongit, line 69: why don't we check the file(s) as we receive them, and then upload them? (instead of upload to s3, then scan)

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -293,9 +342,24 @@ function formatFileSizeMb(bytes, decimalPlaces = 1) {
...
+        super(message);
+        removeFileFromSessionAndDelete(req, file);
+    }
+}
+
+class ErrorAndReloadPage extends Error {
+    constructor(message) {
+        super(`STAYONPAGE: ${message}`);
+    }
+}
```
jenbutongit: this is the same as doing
```
new Error(`STAYONPAGE: ${message}`)
``` 
why is it subclassed?

javascript does not have the same error handling as other languages. In other languages, you're easily able to do a `switch` (or similar) on an error. However, for JS, you need to either do switch on the message, or the `instanceof` (eg `if (instanceof err === ErrorAndReloadPage)` ). 

I think you are only doing a check on the message, instead of instanceof, so there's no need to do this

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -293,9 +342,24 @@ function formatFileSizeMb(bytes, decimalPlaces = 1) {
     return `${(bytes / 1_000_000).toFixed(decimalPlaces)}Mb`;
 }
 
+class ErrorAndDeleteFile extends Error {
+    constructor(message, req, file) {
+        super(message);
+        removeFileFromSessionAndDelete(req, file);
+    }
+}
```
jenbutongit: the rule of thumb is if a function (or class in this case) has "and" in the name it's probably got too many responsibilities. 

Errors should not do anything apart from report the error. For example, what happens when removeFileFromSessionAndDelete fails? why is an error spurting another possible error? 😨 

should go something like this

```.js 
async function addUnsubmittedTag(file, req) {
  try {
  // ...
   await s3.send(new PutObjectTaggingCommand(params));
   sails.log.info(`Only UNSUBMITTED tag added to ${fileStorageName}`);
  } catch (error) {
    throw Error(`STAYONPAGE: ${error}`)
  }
}

try { 
  addCleanAndUnsubmittedTagsToFile(..)
} catch (error) { //this is the STAYONPAGE error. every error from addCleanAndUnsubmittedTagsToFile will return this error

 //optionally check the message/instanceof the error if you want to handle the errors differently

 removeFileFromSessionAndDelete(req, file)
}
```

jenbutongit (changes requested):

api/controllers/FileUploadController.js
RichardBray: I guess my thinking here was more for the errors to be handled in one place but yeah, this doesn't really need to be in the try block.

api/helpers/uploadedFileErrorChecks.js
RichardBray: This is a conversation we had a while back and the reason I brought up using the document-upload tool, or using that node temp folder tool. 

Because multiple large files can be uploaded and we didn't want to store them in memory, Everything gets uploaded to S3 and if they're good, they get a CLEAN tag. If they're not good, they get deleted or automatically deleted after a while because they don't have a CLEAN tag.

I'm happy to create another ticket to look into this personally I prefer the node temp directory option because, no Java. But if we decide it's beneficial for future projects to make changes to the document upload tool then I'll be happy to do it there instead.

api/helpers/uploadedFileErrorChecks.js
RichardBray: I see where you're coming from. This is just my attempt to make the code slightly more readable. I my head it's a bit easier to understand what this is doing
```js
throw new ErrorAndReloadPage(err);
```
more than this:
```js
throw new Error(`STAYONPAGE: ${message}`)
```
But that's all I'm using it for and I don't need the instance benefit but...

I could potentially replace the string includes here with a check for an instance of my custom error type 🤔 
```js
            if (err.message.includes('STAYONPAGE')) { // update to instance of?
                FileUploadController._redirectToUploadPage(res);
            } else {
                res.view('eApostilles/fileUploadError.ejs');
            }
```

api/helpers/uploadedFileErrorChecks.js
RichardBray: This is a really good point I didn't think of this.

I could be completely wrong but what I take from your example is that the function to remove files should be slightly higher up in the chain?

api/helpers/uploadedFileErrorChecks.js
jenbutongit: the main benefit of using it is that most of it is already written, you just need to add some additional settings/code to it so it works for your service. But further down the line, another FCDO service may want to use it and just has to fiddle with env vars here and there.

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -37,48 +44,97 @@ function initialiseClamScan(req) {
...
+        for (const file of req.files) {
+            inDevEnvironment
+                ? await checkLocalFileType(file, req)
+                : await checkS3FileType(file, req);
+        }
+    } catch (err) {
+        if (err.message.includes('not a PDF')) {
+            throw new UserAdressableError(`checkFileType ${err}`);
         }
+        throw new Error(`checkFileType ${err}`);
```
jenbutongit: check slack - it would be better to use error codes rather than doing a string search

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -133,33 +185,32 @@ function getStorageNameFromSession(file, req) {
...
+        await s3.send(new PutObjectTaggingCommand(params));
+        sails.log.info(`Only UNSUBMITTED tag added to ${fileStorageName}`);
+    } catch (err) {
+        throw new Error(`addUnsubmittedTag ${err}`);
+    }
 }
 
 function scanResponses(scanResults, file, req = null, forS3 = false) {
     const { isInfected, viruses } = scanResults;
+    console.log('gets here 3', scanResults.viruses);
```
jenbutongit: do you mean to console.log here?

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -37,48 +44,97 @@ function initialiseClamScan(req) {
...
             port: clamavPort,
         },
     };
-
-    return new NodeClam().init(clamAvOptions);
+    try {
+        return new NodeClam().init(clamAvOptions);
+    } catch (err) {
+        throw new Error(err);
+    }
```
jenbutongit: also check slack for more detail on this. the error will just throw from `NodeClam().init(clamAvOptions);` - so unless you are doing something with the error (like changing the message) there's not too much reason to rethrow it manually. Are you just trying to make it clear that it throws?

jenbutongit (changes requested):

api/helpers/uploadedFileErrorChecks.js
RichardBray: Ah no this was for unit testing, I'll get rid of it :)

api/helpers/uploadedFileErrorChecks.js
RichardBray: Yeah this is a fair point. I've make a ticket to look into doing that (https://consular.atlassian.net/browse/EIA-374) but I don't know how long it will take.

api/helpers/uploadedFileErrorChecks.js
RichardBray: Hmmm, my assumption was that if you use async/await you need to put the await function in a try catch or you'll get an unhandled promise error if there's an error 🤔 

Actually you're right, the parent function is taking care of handling the errors 🤦 
I'll get that fixed.

tests/env.vars
```diff
@@ -0,0 +1,11 @@
+export APPLICATIONDATABASE='{"host": "placeholder","user": "fco_service", "password": "placeholder","database":"placeholder","port": 5432}'
```
jenbutongit, line 1: did you mean to commit this?

api/controllers/FileUploadController.js
```diff
@@ -61,25 +64,54 @@ const FileUploadController = {
...
+
         if (err) {
-            const fileLimitExceeded = err.code === MULTER_FILE_COUNT_ERR_CODE;
+            const fileLimitExceeded =
+                err.code === MULTER_FILE_COUNT_ERR_CODE;
             if (fileLimitExceeded) {
                 req.session.eApp.uploadMessages.fileCountError = true;
             } else {
+                sails.log.error(err);
                 res.serverError(err);
```
jenbutongit: what's the difference between these?

jenbutongit (changes requested):
few comments here - sent over a few things on slack as well 👍

api/controllers/FileUploadController.js
RichardBray: So to my knowledge `res.serverError` sends a 500 error to the client and redirects to the 500 error page, but maybe it logs the error as well so I don't need to log it twice 🤔 
https://sailsjs.com/documentation/reference/response-res/res-server-error

RichardBray: Yep looks like it does log the error 🤦 
https://github.com/UKForeignOffice/loi-application-service/blob/develop/api/responses/serverError.js#L27

tests/env.vars
RichardBray: Yeah basically this file was removed from the whole project during the development of this ticket because it contained some information that was best not public but it's needed to run unit tests locally. And without running unit tests you cannot push code to GH, well that's how I've configured husky. So I've put it back with less information.

jenbutongit (approved):

jenbutongit:
@RichardBray - could you just double check tests/specs/controllers/FileUpload.test.js please? are these tests running? looks like the wrapping test has .skip on it

RichardBray:
> @RichardBray - could you just double check tests/specs/controllers/FileUpload.test.js please? are these tests running? looks like the wrapping test has .skip on it

Hey Jen this is a fair point.

 It's quite a large test file but just the first describe block is skipped, the others should run
Screenshot 2022-02-10 at 15 01 06

The tests I've changed/added should be in the `uploadFileHandler` test block.

PR #339, opened 2022-02-07
Fix issues with automated tests

RichardBray:
# Description
https://consular.atlassian.net/browse/EIA-365
This ticket addresses an issue with our automated a11y testing where it was not picking up issues found by the browser version of axe.

To fix this:
- a new test case was added to check the a11y of the page when a radio button is clicked
- the dependencies involved in this check were upgraded
- custom tags were removed in order to use the default tags that were being blocked

### DIFF
--- a/cypress/integration/a11y.spec.js
+++ b/cypress/integration/a11y.spec.js
@@ -3,9 +3,9 @@ const { findByLabelText, findByRole, findAllByTestId, visit, get, setCookie } = 
 describe('Check accessiblity', () => {
     const TOTAL_ELIBIGILITY_QUESTIONS = 3;
 
-    function checkA11y(options = null) {
+    function checkA11y() {
         cy.injectAxe();
-        cy.checkA11y(options);
+        cy.checkA11y();
     }
 
     function acceptSiteCookies() {
@@ -72,8 +72,6 @@ describe('Check accessiblity', () => {
         });
 
         context('eApp eligibility questions', () => {
-            it('Which service would you like?', () => {});
-
             it('[Error] Which service would you like?', () => {
                 clickContinueBtn();
             });
--- a/cypress/integration/a11y.spec.js
+++ b/cypress/integration/a11y.spec.js
@@ -75,7 +75,7 @@ describe('Check accessiblity', () => {
             });
 
             it('Select radio option and check a11y', () => {
-                findByRole('radio', {name: 'e-Apostille service'}).check();
+                findByLabelText('e-Apostille service').check();
             });
 
             it('Is the e-Apostille accepted in the destination country?', () => {
--- a/cypress/integration/a11y.spec.js
+++ b/cypress/integration/a11y.spec.js
@@ -3,9 +3,9 @@ const { findByLabelText, findByRole, visit, get, setCookie } = cy;
 describe('Check accessiblity', () => {
     const TOTAL_ELIBIGILITY_QUESTIONS = 3;
 
-    function checkA11y() {
+    function checkA11y(options = null) {
         cy.injectAxe();
-        cy.checkA11y(null, {
+        cy.checkA11y(options, {
             runOnly: {
                 type: 'tag',
                 values: ['wcag21aa'],
@@ -90,7 +90,7 @@ describe('Check accessiblity', () => {
 
             it.only('Select radio option and check a11y', () => {
                 findByLabelText('e-Apostille service').check();
-                checkA11y();
+                checkA11y('.govuk-radios__input');
             });
 
             it('Is the e-Apostille accepted in the destination country?', () => {
--- a/cypress/integration/a11y.spec.js
+++ b/cypress/integration/a11y.spec.js
@@ -5,12 +5,7 @@ describe('Check accessiblity', () => {
 
     function checkA11y(options = null) {
         cy.injectAxe();
-        cy.checkA11y(options, {
-            runOnly: {
-                type: 'tag',
-                values: ['wcag21aa'],
-            },
-        });
+        cy.checkA11y(options);
     }
 
     function acceptSiteCookies() {
@@ -22,6 +17,7 @@ describe('Check accessiblity', () => {
     }
 
     beforeEach(() => {
+        cy.clearCookies();
         acceptSiteCookies();
     });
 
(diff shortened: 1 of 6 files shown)


cypress/integration/a11y.spec.js
```diff
@@ -88,6 +83,11 @@ describe('Check accessiblity', () => {
                 checkA11y();
             });
 
+            it('Select radio option and check a11y', () => {
+                findByLabelText('e-Apostille service').check();
+                checkA11y();
```
jenbutongit: worth checking if the button is findable and clickable too

cypress/integration/a11y.spec.js
```diff
@@ -147,7 +147,7 @@ describe('Check accessiblity', () => {
 
         context('eApp applications section', () => {
             function selectFirstApplication() {
-                get('#previousApplications .appRef .govuk-link')
+                get('#dashboard-results tbody .govuk-link')
```
jenbutongit: ideally I would try with testing library selectors first, then add test-ids, and then final resort would be css selectors

cypress/integration/a11y.spec.js
jenbutongit: perhaps checkAlly should be called in `afterEach`?

jenbutongit (changes requested):

khurram-syed-cyb:
The fix for the accessibility issue has been tested and looks fine. There are other issues that have been raised in ticket [EIA-385: Refactoring and fixing automated accessibility testing issues] after the review of accessibility tests. So approving this PR.

khurram-syed-cyb:
I don't have write access in this reop to merge it so @RichardBray LGTM

cypress/integration/a11y.spec.js
RichardBray: Fair point. This is quite a tough case because the item I want to select isn't static, it's sort of dynamic

Screenshot 2022-02-11 at 17 50 54

So I want to select the first link in this table, the first one under Application reference. But the data is retrieved from a db so it's difficult to add an attribute to the element, although.... I could probably do something clever with .ejs code. Let me try.

cypress/integration/a11y.spec.js
RichardBray: Not sure what you mean by `checking if the button is findable and clickable too`. Are you thinking of checking with an if statement to see if the button exists then executing the code if it does?

cypress/integration/a11y.spec.js
jenbutongit: you should use test-ids in that case

cypress/integration/a11y.spec.js
jenbutongit: `cy.findByRole("button").click()`

`findByRole` <- can a screenreader or browser find the button (without an id or css selector)

cypress/integration/a11y.spec.js
```diff
@@ -1,16 +1,11 @@
-const { findByLabelText, findByRole, visit, get, setCookie } = cy;
+const { findByLabelText, findByRole, findAllByTestId, visit, get, setCookie } = cy;
 
 describe('Check accessiblity', () => {
     const TOTAL_ELIBIGILITY_QUESTIONS = 3;
 
-    function checkA11y() {
+    function checkA11y(options = null) {
```
jenbutongit: why default to null? wouldn't it be undefined if nothing is passed in anyway?

cypress/integration/a11y.spec.js
```diff
@@ -36,13 +32,10 @@ describe('Check accessiblity', () => {
             visit('/');
         });
 
-        it('Get your document legalised', () => {
-            checkA11y();
-        });
+        it('Get your document legalised', () => {});
```
jenbutongit: empty test?

cypress/integration/a11y.spec.js
```diff
@@ -79,38 +72,35 @@ describe('Check accessiblity', () => {
...
-            });
+            it('Which service would you like?', () => {});
 
             it('[Error] Which service would you like?', () => {
                 clickContinueBtn();
-                checkA11y();
+            });
+
+            it('Select radio option and check a11y', () => {
+                findByRole('radio', {name: 'e-Apostille service'}).check();
```
jenbutongit: why doesn't `findByLabelText` work here?

cypress/plugins/index.js
```diff
@@ -16,7 +16,7 @@
  * @type {Cypress.PluginConfig}
  */
 // eslint-disable-next-line no-unused-vars
-module.exports = (on, config) => {
+module.exports = (_on, config) => {
   config = require('cypress-dotenv')(config);
```
jenbutongit, line 20: you can move 
```
        cy.on('fail', () => {
             cy.screenshot();
         });
```
to here, so it happens for all your test files, not just a11y spec.

jenbutongit (changes requested):

cypress/integration/a11y.spec.js
RichardBray: So I read your response to this question https://github.com/UKForeignOffice/loi-application-service/pull/339#discussion_r805421555

As the role should be used to check if the element is findable, but I could be wrong.

cypress/plugins/index.js
RichardBray: I tried this but it looks like `fail` isn't an option that can be used in that file.

Screenshot 2022-02-21 at 14 52 49

Unless I'm doing something wrong.

cypress/integration/a11y.spec.js
jenbutongit: for form inputs/labels, they recommend findByLabelText since this is how users (screenreader/otherwise) navigate through a form 
https://testing-library.com/docs/queries/about/#priority

cypress/plugins/index.js
jenbutongit: np - feel free to resolve. does on.fail('..') work in your file?

jenbutongit (approved):

cypress/plugins/index.js
RichardBray: Yeah the cy.on('fail', ()... does work, although I do prefer to click though/inspect the issue to see what actually went wrong so I don't use it that much 🤷

PR #346, opened 2022-02-23
Feat/eia 385 refactor a11y tests

RichardBray:
# Description
Refactor tests to prevent logging in and running through the same steps for each tests.

### Before
Screenshot 2022-02-23 at 16 22 19

### After
Screenshot 2022-02-23 at 16 04 05

### DIFF
--- a/cypress/integration/a11y.spec.js
+++ b/cypress/integration/a11y.spec.js
@@ -133,7 +133,7 @@ describe('Check accessiblity', () => {
             checkA11y('eApp file upload');
 
             uploadTestFile();
-            cy.contains(' 1 file was uploaded ')
+            findByText('1 file was uploaded', { exact: false });
             checkA11y('Add your PDFs - 1 file uploaded');
 
             clickContinueBtn();
--- a/cypress/integration/a11y.spec.js
+++ b/cypress/integration/a11y.spec.js
@@ -1,9 +1,15 @@
-const { findByLabelText, findByRole, findAllByTestId, visit, get, setCookie } =
-    cy;
+const {
+    findByLabelText,
+    findByRole,
+    findAllByTestId,
+    visit,
+    get,
+    setCookie,
+    go,
+    wait,
+} = cy;
 
 describe('Check accessiblity', () => {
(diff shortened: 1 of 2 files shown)


cypress/integration/a11y.spec.js
```diff
@@ -69,114 +93,68 @@ describe('Check accessiblity', () => {
...
+            go('back');
+            checkRadioAndClickContinue('Yes');
+            clickContinueBtn();
+
+            visit('/upload-files');
+            checkA11y('eApp file upload');
+
+            uploadTestFile();
+            wait(500);
+            checkA11y('Add your PDFs - 1 file uploaded');
```
jenbutongit: is it necessary to wait? cypress waits by itself

jenbutongit (changes requested):

cypress/integration/a11y.spec.js
RichardBray: Ah didn't know cypress waits by itself thanks for pointing that out. I tried your suggestion in slack to use `findByText` but it didn't seem to work:

Screenshot 2022-02-24 at 16 39 17

I think it's because that text is an .ejs conditional and doesn't show up on initial page load 🤷 

But `contains` seems to work better:

Screenshot 2022-02-24 at 17 50 25

jenbutongit: cypress runs on the browser, so it will just be rendered html. ejs is server side rendered as well, so ejs shouldn't interfere. I can't tell if it's the styling, but it looks like for findByText, this was entered ` 1 file was uploaded ` (whitespace at the start and end). you can use `findByText("thing I want to search", { exact: false })`. (but this is minor and contain works fine also)

jenbutongit (approved):

cypress/integration/a11y.spec.js
RichardBray: Ah you're right about the exact false flag because the text is actually on a new line in the code so maybe it's looking for exact whitespacing as well. I'll give that a go tomorrow and see if that fixes it.

PR #350, opened 2022-03-01
Replace request files with session data to fix race condition

RichardBray:
# Description

There was a bug with the app which, I'm surprised wasn't caught before where, if a large file is uploaded then automatically deleted by the code, the file type check expects it to be there even though it as been deleted.

I believe this happens because the file type check happens before the file is actually deleted:

Screenshot 2022-03-01 at 18 01 15

To fix this I've changed the code to check for files from the session data instead of the request data (req.files) because the session data updates a lot faster, since it's not waiting for an actual file to delete.

Open to suggestions on how to make this better 👍 

## Actually...

Now that I've had some time to think about this I'm not sure if it's a race condition or just my lack of understanding in how `req.files` works. 

I do have a deleteFileHandler function which works but I don't know if the site requires a full page refresh for that to work 🤷 

Happy to explain this PR in details on request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,3 +162,4 @@ uploads
 test-results.xml
 a11y-test-results.xml
 cypress/screenshots
+cypress/browsers/*/*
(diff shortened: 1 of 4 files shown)


PaulioRandall (approved):
Reviewed through remote walkthrough.

PR #352, opened 2022-03-04
feat: reset upload errors on different pages

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-411

This PR is to reset file upload errors when the user leaves the page. 

Before if there was an error on the document upload page like the one below:

Screenshot 2022-03-01 at 15 27 28

And the user navigates back then revisits the upload page, the error still persists. This PR is here to address that issue.


api/controllers/FileUploadController.js
```diff
@@ -61,11 +61,13 @@ const FileUploadController = {
...
-        req.session.eApp.uploadMessages.infectedFiles = [];
-        req.session.eApp.uploadMessages.noFileUploadedError = false;
+    clearExistingErrorMessages(req) {
+        req.session.eApp.uploadMessages = {
+            errors: [],
+            infectedFiles: [],
+            fileCountError: false,
+            noFileUploadedError: false,
+        }
     },
```
jenbutongit: instead of doing this manually, you probably actually want to use a feature like "flash" for this, sorry I didn't catch this before.

it's a fairly common server decoration/utility regardless of which framework you use (i.e. hapi/sails/express). You use it for showing short-lived messages. Typically, as soon as the message is read, it will delete itself or is deleted on the next request. 

It would work like this
1. user uploads erroneous file  (POST)
2. server detects error
3. server uses req.flash to store the error message(s)
4. response redirects user to the page with the error (REDIRECT)
5.  request is now GET. In the get request, if it can retrieve a message from req.flash, display it

Post-Redirect-Get is a common and well documented pattern, and using req.flash suits it quite well. 

unfortunately as of sails v1, they've removed the flash middleware and you have to install it yourself. https://sailsjs.com/documentation/upgrading/to-v-1-0#other-breaking-changes. they suggest this: https://www.npmjs.com/package/connect-flash

jenbutongit (changes requested):

api/controllers/FileUploadController.js
```diff
@@ -23,6 +23,12 @@ const FileUploadController = {
         const connectedToClamAV = await connectToClamAV(req);
         // @ts-ignore
         const userData = HelperService.getUserData(req, res);
+        const noFileUploadedError = Boolean(
+            req.flash('noFileUploadedError')[0]
+        );
+        const fileCountError = Boolean(req.flash('fileCountError')[0]);
+        const errors = req.flash('errors');
+        const infectedFiles = req.flash('infectedFiles');
```
RichardBray: Putting the flash messages in variables seemed to retain them for me. Without it they'd disappear and not show on the page. Unless I call flash directly form the .ejs page.

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -291,32 +284,15 @@ function checkTypeSizeAndDuplication(req, file, cb) {
...
-                    ...errors,
-                ];
-            }
-        });
-    } else {
-        req.session.eApp.uploadMessages.errors.push({
-            filename: fileName,
-            errors,
-        });
-    }
```
RichardBray, line 319: This code was here, well the first part of the if statement, because there was an assumption that one file could have multiple errors, so there was a check to add an error to an already existing file and display it to the user, sort of like bullet points.

But I've noticed a file will always usually have one error. And if it does have one error (not a pdf, infected with virus, too large, duplicate), the file is usually removed

api/helpers/uploadedFileErrorChecks.js
```diff
@@ -291,32 +284,15 @@ function checkTypeSizeAndDuplication(req, file, cb) {
...
+function addErrorsToPage(req, filename, errors) {
+    req.flash('errors', [{ filename, errors }]);
```
jenbutongit: is this necessary? 

to add an error to a page, you would just do req.flash('errors', ['myErrors']). I think also you've used a mixture of addErrorsToPage and req.flash in some places, which may be confusing things. Stick to one - I would just go with req.flash

api/controllers/FileUploadController.js
```diff
@@ -23,6 +23,12 @@ const FileUploadController = {
         const connectedToClamAV = await connectToClamAV(req);
         // @ts-ignore
         const userData = HelperService.getUserData(req, res);
+        const noFileUploadedError = Boolean(
+            req.flash('noFileUploadedError')[0]
+        );
+        const fileCountError = Boolean(req.flash('fileCountError')[0]);
```
jenbutongit: why not `req.flash('errors', ['my new error'])`, and every time there's a new error, you keep adding to `req.flash('error', [..])`? 

then, when you render, all you do is `errors: req.flash('errors')` 

also, Boolean check here is confusing. 

You can do `req.flash('errors', ['noFileUploadedError'])` as a string. Then, inside your template, you can do 
`if(errors.find('noFileUploadedError'))`

jenbutongit (changes requested):

api/helpers/uploadedFileErrorChecks.js
RichardBray: Good point

api/controllers/FileUploadController.js
```diff
@@ -23,6 +23,9 @@ const FileUploadController = {
         const connectedToClamAV = await connectToClamAV(req);
         // @ts-ignore
         const userData = HelperService.getUserData(req, res);
+        const flashErrors = req.flash('errors');
+        const noBooleanErrors = flashErrors.filter(error => error !== 'fileCountError' && error !== 'noFileUploadedError');
+        const infectedFiles = noBooleanErrors.filter(error => typeof error === 'string');
```
jenbutongit: I think instead of doing multiple checks and filtering here, you can just use another flash key for the infected files. 
`req.flash('infectedFiles', [...])`. then you don't need to check for noBooleanErrors and infectedFiles via a filter.

api/controllers/FileUploadController.js
```diff
@@ -61,34 +69,28 @@ const FileUploadController = {
...
         }
 
-       removeFilesIfLarge(req);
+        removeFilesIfLarge(req);
 
         if (err) {
             const fileLimitExceeded = err.code === MULTER_FILE_COUNT_ERR_CODE;
             if (fileLimitExceeded) {
-                req.session.eApp.uploadMessages.fileCountError = true;
+                req.flash('errors', ['fileCountError']);
```
jenbutongit: you could use the codes for req.flash. eg `req.flash('errors', [err.code])`. It will keep the errors synced up as well, so you don't have to manage separate names for errors (codes and string).

 Although it works better with typescript + enums with intellisense.

jenbutongit (changes requested):

api/controllers/FileUploadController.js
RichardBray: yeah that's true

RichardBray: Ha, looks like I forgot to update [this line](https://github.com/UKForeignOffice/loi-application-service/pull/352/files#diff-ed36c26b88984142f9082154426401bd5611ae8a2e1d593e7863b9eb54a25da5R209) so my refactored code would not have worked anyway 🤦 

But it makes it easier for me to update the code

api/controllers/FileUploadController.js
RichardBray: hmm, does this make sense if I'm only handling one error code?

views/eApostilles/uploadFiles.ejs
```diff
@@ -10,79 +14,44 @@
...
-                    <li>
-                        <a href="#file-upload-form">No files have been selected</a>
-                    </li>
+                    <% if (preUploadErrorsExists) { %>
+                        <li>
+                            <a href="#error-list">One or more files could not be uploaded</a>
+                        </li>
+                    <% } %>
+                    <% if (postUploadErrorsExists) { %>
+                        <% postUploadErrors.forEach((postUploadError) => { %>
```
jenbutongit: no need to action, more of a code style/cleaniness thing - I would just default to an empty array in `api/controllers/FileUploadController.js`, so you don't have to do length check here. Template files generally include a lot of if statements and can get hard to read

views/eApostilles/uploadFiles.ejs
```diff
@@ -10,79 +14,44 @@
...
+                            <li>
+                                <a href="#file-upload-form"><%= postUploadError %></a>
+                            </li>
+                        <% }); %>
+                    <% } %>
+                    <% if (infectedFileExists) { %>
+                        <li>
+                            <% if (infectedFiles.length === 1) { %>
+                                <a href="#">
```
jenbutongit: href to nowhere?

jenbutongit (changes requested):
ooh. getting so clean 🧼 - just a couple things

views/eApostilles/uploadFiles.ejs
RichardBray: ball! Good spot, I'll fix that

jenbutongit (approved):
🪄

PR #353, opened 2022-03-14
chore: add more checks for accessiblity

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-385

This ticket is to address a few things about the accessibility tests raised by Khurram. 

These are all focused around the 3 eligibility pages in the eApp flow:

* Checking the accessibility on errors specific to the three eligibility questions
* Checking accessibility after a radio button has been selected
* Checking the accessibility of the exit pages, when someone selects 'No' the presses 'Continue'

Screenshot 2022-03-15 at 17 12 57

Screenshot 2022-03-15 at 17 13 02

Screenshot 2022-03-15 at 17 13 09

### DIFF
--- a/cypress/integration/a11y.spec.js
+++ b/cypress/integration/a11y.spec.js
@@ -87,7 +87,7 @@ describe('Check accessiblity', () => {
                 '3 - Have the PDFs been notarised and digitally signed by a notary?',
             ];
 
-            for (const page of eligibilityPages) {
+            for (const [index, page] of eligibilityPages.entries()) {
                 checkA11y(page);
 
                 clickContinueBtn();
@@ -100,6 +100,9 @@ describe('Check accessiblity', () => {
                 }).click();
 
                 findByLabelText('Yes').check();
+
+                if (index === 2) findByTestId('prepare-pdf').click();
+
                 checkA11y(`${page}- Radio Check`);
                 clickContinueBtn();
             }
--- a/cypress/integration/a11y.spec.js
+++ b/cypress/integration/a11y.spec.js
@@ -87,7 +87,7 @@ describe('Check accessiblity', () => {
                 '3 - Have the PDFs been notarised and digitally signed by a notary?',
             ];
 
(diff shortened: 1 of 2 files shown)


cypress/integration/a11y.spec.js
```diff
@@ -80,6 +80,34 @@ describe('Check accessiblity', () => {
...
         });
 
+        function eligibilityPages() {
+            const eligibilityPages = [
+                '1 - Is the e-Apostille accepted in the destination country?',
+                '2 - Check if the documents are eligible for the e-Apostille service',
+                '3 - Have the PDFs been notarised and digitally signed by a notary?',
+            ];
+
+            for (const [index, page] of eligibilityPages.entries()) {
```
jenbutongit: you can just do 
```
  eligibilityPages.forEach((page, index) => {...}) 
```

fyi, `.map`, `.filter`, `.forEach` etc will give you the item that is being iterated on (page in this case) and the index. the 3rd parameter is usually the array being iterated over `eligibilityPages` 

however I would just avoid this abstraction entirely. what was wrong with it before?

jenbutongit (changes requested):

cypress/integration/a11y.spec.js
RichardBray: Nothing major just a lot of repeating since the three eligibility pages need the exact same checks.

Well actually, the issue was each page only had 1 check before and now they have 4, which was starting to look like a lot.

jenbutongit: tests should be as simple as possible IMO. It makes them easier to maintain, and split apart if need be. Could you not just do this iteration/function inside the `it` block?

jenbutongit: and Roger on forEach - actually I think a for loop would be better since you're not dealing with mutation.

jenbutongit (changes requested):

jenbutongit (approved):

khurram-syed-cyb:
LGTM

PR #354, opened 2022-03-18
Feat/eia 435 remove eapp flag

RichardBray:
# Description

This PR is to remove the `electornicEnabled` flag from the app and remove the column from the database so that anyone can have access to the eApp.

- Database migration changes have been applied locally and work but won't be applied on AWS until the PR has been merged and Khurram has done his testing

- The eApp specific dashboard has been removed and the code from that page has been added to the proper dashboard

- The eApp specific service selection page has not been replaced as that will be done in a future ticket


api/controllers/DashboardController.js
```diff
@@ -125,19 +120,17 @@ const DashboardController = {
             });
     },
 
-    _chooseStoredProcedure(secondarySortOrder, electronicEnabled = false) {
-        const procedureToUse = electronicEnabled
-            ? 'dashboard_data_eapp'
-            : 'dashboard_data';
+    _chooseStoredProcedure(secondarySortOrder) {
+        const procedureToUse = 'dashboard_data_eapp';
```
alisalaman-cyb, line 124: me being pedantic but if consolidating these, wouldn't you not include the _eapp suffix and keep the more generic dashboard_data name?

database_change_log/EIA-435_depete_eApp_flag.sql
```diff
@@ -0,0 +1,3 @@
+-- Run against FCO-LOI-Users
+
+ALTER TABLE "Users" DROP COLUMN  "electronicEnabled";
```
alisalaman-cyb, line 3: This script should only be executed after a period of use (1 or 2 weeks) in prod of the new functionality.  This way you have an easier rollback plan should you need to revert the codebase to the previous version in prod.  This should be explicitly stated in release notes provided to Kainos when requesting the release.

views/partials/dashboardResults.ejs
```diff
@@ -1,103 +1,158 @@
...
-      <% } %>
-    </tr>
-    <% } %>
+        <% for (i=0;i<users_applications.length;i++) { %>
+            <tr style="vertical-align: text-top">
+                <td class="appDate">
+                    <%= dayjs(users_applications[i].createdDate).format('DD-MM-YYYY') %>
+                </td>
+                <td>
+                    <% if (users_applications[i].applicationtype === 'e-Apostille') { %>
```
alisalaman-cyb: You could probably replace this bloaty if statement with a getUrlPath function on the users_application object.  The  if statement could then be contained within there allowing you to replace this block with something like (or equivalent syntax for inline variables)

`....
href="<%= users.applications[i].getUrlPath/>">
....`

views/partials/dashboardResults.ejs
```diff
@@ -1,103 +1,158 @@
...
+            </tr>
+        <% } %>
     </tbody>
 </table>
 
 
 
 <div class="pager">
-    <p class="pager-summary"><%= paginationMessage %></p>
+    <p class="pager-summary">
```
alisalaman-cyb: Not sure if possible in this repo but you may want to consider creating a fragment for the pagination details so you can just insert a reference / import to it here.  It'll then make the dashboardResults page more specific to the result data rather than the boilerplate stuff like pagination functionality.

alisalaman-cyb (commented):
only a few minor things to query/consider.  Note, the point about using the generic variable name rather than the "eApp" suffixed ones applies to a number of places but I've only made reference to it once.  Again, just something to consider there.

api/controllers/DashboardController.js
RichardBray: Hey Ali, this is a really fair point, I left it because i thought it was a bit of a faff to change stored functions in the DB but it does make sense to change.

database_change_log/EIA-435_depete_eApp_flag.sql
RichardBray: This makes sense, I'll make a not of it in this file and when Kainos review this PR again I'll mention it to them

api/controllers/DashboardController.js
alisalaman-cyb: ah ok, I overlooked that these were stored procedures.  Yep, agreed in that case, too much of a faff to change this particular area but there are also other areas in regular JS code that you may want to consider removing the eApp suffix on in a number of areas throughout.

views/partials/dashboardResults.ejs
RichardBray: Hey @alisalaman-cyb when you say fragment I'm assuming you mean a separate html page (possibly with a separate controller) for the pagination?

alisalaman-cyb: yeah pretty much.  Essentially it's a way to extract out a block of logic / functionality to a separate file so you're left with a single "import" line in the parent which allows you to just focus on the business logic rather than perhaps boilerplate code like headers and footers.  I found this example of using partials for EJS template engine which it looks like this app is using: https://www.digitalocean.com/community/tutorials/how-to-use-ejs-to-template-your-node-application

api/controllers/OpenPaperAppController.js
```diff
@@ -1,6 +1,6 @@
 const SummaryController = require('./SummaryController');
 
-const OpenAppController = {
+const OpenPaperAppController = {
```
alisalaman-cyb, line 3: Feels like this should be OpenAppController?

views/partials/dashboardResults.ejs
alisalaman-cyb: Resolved with the inclusion of the viewAppUrl property

api/controllers/DashboardController.js
alisalaman-cyb: FYI, I made a reference to such a case where you're still using OpenPaperAppController but in the new state I think OpenAppController would probably be a better name.

api/controllers/OpenPaperAppController.js
RichardBray: Fair point, the reason this was names is because the paper and eApps have different views from the dashboard.

This is what an eApp looks like:
Screenshot 2022-03-29 at 17 37 20


And this is what a paper app looks like:

Screenshot 2022-03-29 at 17 40 30

We have plans to change them to make them more similar but now they look quite different.

alisalaman-cyb:
@RichardBray as discussed LGTM

PR #357, opened 2022-03-31
Feat/eia 426 move choose service page

RichardBray:
# Description
This PR is focused on moving the eApp specific service selection screen which was only accessible after logging in, to the pre-login part of the application.

Current pre-login service selection screen:

Screenshot 2022-03-31 at 14 23 59

New pre-login service selection screen:

Screenshot 2022-03-31 at 14 11 56

This PR also contains:
- Updated tests for the specific feature
- Redirecting to a different place if eApp is selected pre-login since login is still required for eApp
- Rendering an error page if no radio option is selcted


api/controllers/ApplicationTypeController.js
```diff
@@ -54,93 +55,87 @@ module.exports = {
...
+    },
+
+    _renderServiceSelectionPage(req, res, userModels) {
+        const userLoggedIn = HelperService.LoggedInStatus(req);
+        const userData = HelperService.getUserData(req, res);
+        const pageData = {
             application_id: 0,
             userServiceURL: sails.config.customURLs.userServiceURL,
-            error_report: false,
+            error_report: Boolean(req.query.error),
```
RichardBray: This is the main bit that was changed, the rest is refactoring for testing reasons.

api/controllers/ApplicationTypeController.js
```diff
@@ -54,93 +55,87 @@ module.exports = {
...
+            back_link: '/',
         });
+    },
 
+    _addUserAccountToSession({req, res, userModels, pageData}) {
+        return userModels.User.findOne({where: {email: req.session.email}}).then((user) => {
+            return userModels.AccountDetails.findOne({where: {user_id: user.id}}).then((account) => {
+                const standardAppCountQuery = 'SELECT count(*) FROM "Application" WHERE "user_id" =:userId and "serviceType" = 1 and "createdAt" > NOW() - INTERVAL \'' + sails.config.standardServiceRestrictions.appSubmissionTimeFrameInDays + ' days\' and ("submitted" =:submitted OR "submitted" =:queued)';
+
+                console.log('this gets hit 3', account)
```
jenbutongit: remove

api/controllers/ApplicationTypeController.js
```diff
@@ -54,93 +55,87 @@ module.exports = {
...
+                return sequelize.query(
+                    standardAppCountQuery,
+                    { replacements: {userId: user.id, submitted: 'submitted', queued: 'queued'},
```
jenbutongit, line 100: submitted: true? are there other states?

api/controllers/ApplicationTypeController.js
```diff
@@ -54,93 +55,87 @@ module.exports = {
...
+                    type: sequelize.QueryTypes.SELECT
+                }).then(() => {
+                    req.session.user = user;
+                    req.session.account = account;
+                    req.session.email_sent = false;
+                    console.log('this gets hit 4', {
```
jenbutongit: remove

views/applicationForms/applicationType.ejs
```diff
@@ -1,76 +1,181 @@
...
+                            <div class="govuk-!-padding-left-3">
+                                <p>
+                                    Same day London drop-off and collection for businesses only
+                                </p>
+                                <ul class="govuk-list govuk-list--bullet govuk-list--spaced">
+                                    <li>£75 per document</li>
+                                    <li>Each application is limited to 20 documents</li>
+                                    <li>The service is open on working days, 9am to 12:30pm for drop-off</li>
+                                    <li>A collection time will be allocated between 1:30pm to 4:30pm</li>
+                                    <li>To use the premium service you need to create an account. You'll be asked to provide company details</li>
```
jenbutongit: GitHub syntax highlighting seems to think this line onwards is broken html. do you mind just double checking?

api/controllers/ApplicationTypeController.js
jenbutongit: should this be using post redirect get + flash?

jenbutongit (changes requested):

api/controllers/ApplicationTypeController.js
RichardBray: Yeah I guess this could be using a flash but I'm not sure about post redirect get? I guess I'm not quite sure what you mean. 😅

api/controllers/ApplicationTypeController.js
RichardBray: Yep there's 'queued', 'draft' and false. Would make sense to have this as a frozen object/enum somewhere actually so that other's don't get confused.

Happy to create a follow up refactor ticket to look into this in the future.

views/applicationForms/applicationType.ejs
RichardBray: LGTM
Screenshot 2022-04-01 at 14 34 12

tests/specs/controllers/ApplicationType.test.js
```diff
@@ -25,6 +25,10 @@ var testApplicationId = 8072;
 var testEmail = 'mark.barlow@digital.fco.gov.uk';
 var testApplication_id = 7888;
 
+function assertWhenPromisesResolved(assertion) {
```
jenbutongit: does the test framework you use not await?

views/applicationForms/applicationType.ejs
jenbutongit: all good. probably just doesn't like ejs highlighting ¯\_(ツ)_/¯

api/controllers/ApplicationTypeController.js
```diff
@@ -54,93 +55,86 @@ module.exports = {
...
+    _renderServiceSelectionPage(req, res, userModels) {
+        const userLoggedIn = HelperService.LoggedInStatus(req);
+        const userData = HelperService.getUserData(req, res);
+        const errorReport = req.flash('serviceSelectError')[0];
+
+        const pageData = {
             application_id: 0,
             userServiceURL: sails.config.customURLs.userServiceURL,
-            error_report: false,
+            error_report: Boolean(errorReport),
```
jenbutongit: `error_report: req.flash('serviceSelectError')` ideally should return the error message "You must select a service type." - at the moment it appears that there can only be one error, however by passing the string through it makes it much more flexible

jenbutongit (changes requested):

tests/specs/controllers/ApplicationType.test.js
RichardBray: hmmm, fair point, I will try that

api/controllers/ApplicationTypeController.js
```diff
@@ -54,93 +55,87 @@ module.exports = {
...
+        const userLoggedIn = HelperService.LoggedInStatus(req);
+        const userData = HelperService.getUserData(req, res);
+        const errorMessage = String(req.flash('serviceSelectError'));
+
+        const pageData = {
             application_id: 0,
             userServiceURL: sails.config.customURLs.userServiceURL,
-            error_report: false,
+            errorMessage,
+            errorMessageExists: Boolean(errorMessage),
```
jenbutongit: you can just do a !errorMessage check in the template

views/applicationForms/applicationType.ejs
```diff
@@ -1,76 +1,177 @@
...
-            Telephone: 03700 00 22 44 <br>
-            Monday to Friday, 9am to 5pm<br>
-            <a class="govuk-link" href="https://www.gov.uk/call-charges" rel="external"  target="_blank">Find out about call charges <span class="visually-hidden"> opens in new window</span></a>
-        </p>
+    <div class="column-two-thirds">
+        <div class="govuk-inset-text govuk-!-margin-top-0">
+            It takes up to 2 working days to process your documents, plus courier or postage time if you choose the standard (paper-based) service.
         </div>
+        <form
+            action="/handle-service-choice"
```
jenbutongit: this should be the same URL as the get request, not a separate route.

GET /select-service -> renders the page
POST /select-service -> handles the post request

jenbutongit (changes requested):

api/controllers/ApplicationTypeController.js
RichardBray: This is true, but it still needs to be converted to a boolean in the template. 

By default error message is `[]` and when there's an error message in it, then it returns a string. And as I prefer `Boolean(errorMessage)` to `!! errorMessage`, I might just do that in the template 😁

jenbutongit (approved):

PR #359, opened 2022-04-01
Feat/eia 432 move suitability questions

RichardBray:
# Description
https://consular.atlassian.net/browse/EIA-432

This PR is similar to and closely linked to the [move service page PR](https://github.com/UKForeignOffice/loi-application-service/pull/357). 

This entails moving the suitability questions from the post login stage of the eApp flow to the pre-login stage in order for eApp users to start the process of creating an eApp before they signing in so that they know if an eApp is the right option for them.

## Flow before this PR

https://user-images.githubusercontent.com/1377253/161528590-d63eccd3-decf-4c3d-9b5f-866317b10a60.mp4

## Flow after this PR

https://user-images.githubusercontent.com/1377253/161528618-39bac133-c100-4344-8333-e6154ff84d15.mp4


api/controllers/ApplicationTypeController.js
```diff
@@ -54,93 +54,90 @@ module.exports = {
...
+        return res.view('applicationForms/applicationType.ejs', {
+            ...pageData,
+            back_link: '/',
+        });
     },
 
-    handleServiceChoice(req, res) {
-        const {'choose-a-service': chosenService} = req.body;
-        const userLoggedIn = HelperService.LoggedInStatus(req);
+    async _addUserAccountToSession({req, res, userModels, pageData}) {
```
alisalaman-cyb, line 94: this function doesn't appear to persist any data, I can only see queries here.  Can you clarify where the user session data gets persisted?

api/controllers/ApplicationTypeController.js
```diff
@@ -54,93 +54,90 @@ module.exports = {
...
-                error_report: true,
-                user_data: HelperService.getUserData(req, res),
-                back_link: req.session.startBackLink,
-            });
+            req.flash('serviceSelectError', 'You must select a service type.');
+
+            return res.redirect('/select-service');
         }
-        return res.redirect(servicePages[chosenService]);
+        return res.redirect(servicePages[chosenService] || servicePages.default);
```
alisalaman-cyb, line 141: this OR condition looks like it's redundant as servicePages will always return servicePages.default if none of the values match?

api/controllers/ApplicationTypeController.js
```diff
@@ -314,51 +314,50 @@ module.exports = {
...
+                                    }
+
+                                    const UserModels = getUserModels(
+                                        req._sails.config
+                                            .userServiceSequelize
+                                    );
+                                    const userDataFromDB = await UserModels.User.findOne({where:{email:req.session.email}});
+                                    const accountDetailsFromDB = await UserModels.AccountDetails.findOne({where:{user_id:userDataFromDB.id}});
+
+                                    await UsersBasicDetails.create({
```
alisalaman-cyb: The user is created inside a callback after a call to User.findOne and AccountDetails.findOne.  Is this still needed given you're dependent on valid values from the accountdetailsFromDB and userDataFromDB?

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -17,6 +21,32 @@ const CheckUploadedDocumentsController = {
         });
     },
 
+    async _addUserDataToDB(req) {
```
alisalaman-cyb: This block looks like it's duplicated in ApplicationTypeController.newApplication.  Consider separating it out to a function in a helper class that can then be referenced in both controllers.

api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -17,6 +21,32 @@ const CheckUploadedDocumentsController = {
...
+            application_id: req.session.appId,
+            first_name: accountDetailsFromDB.first_name,
+            last_name: accountDetailsFromDB.last_name,
+            telephone: accountDetailsFromDB.telephone,
+            mobileNo: accountDetailsFromDB.mobileNo,
+            email: userDataFromDB.email,
+            confirm_email: userDataFromDB.email,
+            has_email: true,
+        }
+        await UsersBasicDetails.create(userDetails);
```
alisalaman-cyb: What happens if the create fails?  It maybe covered in the create function but it looks like it needs error handling here.

api/controllers/EAppEligibilityQuestionsController.js
```diff
@@ -29,18 +29,10 @@ const EAppEligibilityQuestionsController = {
 
     _fetchUserData(req, res) {
         const userData = HelperService.getUserData(req, res);
```
alisalaman-cyb: Can you condense this to just `return HelperService.getUserData(req, res);`

views/applicationForms/applicationType.ejs
```diff
@@ -1,76 +1,181 @@
...
-                <li>Milton Keynes drop-off and collection only</li>
-                <li>overnight service</li>
-                <li>the counter opens on working days from 1pm to 2pm</li>
+            </div>
+        <%} %>
+    </div>
 
-            </ul>
-            <a class="govuk-button govuk-button--secondary" id="dropoff-service" href="/new-application?app_type_group=3">Start a business drop-off application</a>
+    <%if (errorMessageExists) { %>
```
alisalaman-cyb, line 19: something to consider: split the error message block out into a fragment / partial so you can import it in any page where you need an error message to be displayed RATHER THAN using the HTML for the error block in each page.  You can then have a single line where you pass the error variable (error message) to the fragment and your page will contain just the page-specific information rather than boilerplate guff.

views/applicationForms/applicationType.ejs
```diff
@@ -1,76 +1,181 @@
...
+                    </div>
 
-</div>
+                    <details class="govuk-details" data-module="govuk-details">
+                        <summary class="govuk-details__summary">
+                            <span class="govuk-details__summary-text">
+                                How much are the courier fees or postage?
+                            </span>
+                        </summary>
+                        <div class="govuk-details__text">
```
alisalaman-cyb, line 103: Not sure how much this information changes but may want to consider splitting the courier variables into config so you can change them in the config and they get applied in whatever pages need them.

alisalaman-cyb (commented):
A few things to look at and verify.

api/controllers/ApplicationTypeController.js
RichardBray: I can confirm the session data does get persisted. SailsJS is a little weird like that, there's a lot of magic happening in the background.

Basically `req.session` is configured to update to use a redis database as a session. So on this line:
```js
req.session.account = accountDataFromDB
```
Updates the account entry in the redis database, but yeah it is a bit weird.

api/controllers/ApplicationTypeController.js
RichardBray: Maybe, but I'm not sure if it will. `servicePages ` is an object with 5 keys and one of them is `default`. 

If none of the keys values are matched, `servicePages ` will just return `undefined`. 

Screenshot 2022-04-04 at 17 37 34

So the `|| servicePages.default` is supposed to prevent the site from breaking on the user. Maybe I haven't written the code in the most readable way.

api/controllers/ApplicationTypeController.js
RichardBray: Hey Ali, I'm not quire sure I understand the question. 

1. Is it more to do with positioning? The create could run before the two variables above it others and could cause a race condition since it needs information from them?

2. Or more for necessity. It the create needed at all and can it be removed?

---

If your question is the first. To my understanding then the JS code is read line by line, the await conditions will need to complete before the code moves on to the next line. So in theory, there shouldn't be a race condition. But if for some reason there is, it should be caught by the catch function on line **362**.

For the second question. I'd admit it is a bit weird to create a new user when the user is creating the application especially if the user already has details saved in the DB. I had a chat with Kainos about this and that's how they received the service. I think for a standard paper application, this makes sense because you can create one of these without an account. But for the other types of applications, e-App included, you need an account.

It's not very easy to follow but on line **326** eApps have an appType of 4, so they wouldn't hit any of this code, a user is created later on in the flow for eApps.

views/applicationForms/applicationType.ejs
RichardBray: That's a good point. I'll check with Kainos first. They tend to populate the config variables between AWS environments since I don't have access to all of them.

api/controllers/ApplicationTypeController.js
alisalaman-cyb: Ah my bad.  The await should cover it as the function calls will effectively get called synchronously.

api/controllers/ApplicationTypeController.js
alisalaman-cyb: Got it.  Thanks for clarifying

api/controllers/ApplicationTypeController.js
alisalaman-cyb: thanks for explaining.  Very magic!

api/controllers/CheckUploadedDocumentsController.js
RichardBray: This is a fair point. I was practicing a bit of [AHA programming](https://kentcdodds.com/blog/aha-programming) with this one but it's not that big of a deal to change.

views/applicationForms/applicationType.ejs
RichardBray: This is a good idea. As this is a bit of a bigger change involving multiple pages on the site, (if it were to be done properly that is). Would it make sense to create a separate ticket for this to be done later?

views/applicationForms/applicationType.ejs
RichardBray: According to Kainos, these values change so rarely it won't make that much of a different to put them in the config. 
However, it might make more sense to read the prices from the `PostagesAvailable ` table in the DB.

As I don't have any experience with that table, are you happy for me to create a different ticket for this specific change?

alisalaman-cyb: No problem.  It was just a thought so it sounds like there's actually not much benefit in making the change - your call to raise a separate ticket or scrap it.

views/applicationForms/applicationType.ejs
alisalaman-cyb: yeah sure.

api/controllers/EAppEligibilityQuestionsController.js
```diff
@@ -28,17 +28,7 @@ const EAppEligibilityQuestionsController = {
...
-
-        return userData;
-    },
-
-    _checkUserLoggedIn(userData, res) {
-        if (!userData.loggedIn) {
-            sails.log.error('User is not logged in', userData);
-            return res.forbidden();
-        }
+        return HelperService.getUserData(req, res);
```
alisalaman-cyb: I think I may have missed something with my previous comment.  It looks like it needs to use the user data to check if they're logged in so this probably needs to be reinstated as this validation check is now removed with your latest change

alisalaman-cyb (commented):
one minor change to revert the check in EAppEligibilityQuestionsController.  The rest looks good.

api/controllers/EAppEligibilityQuestionsController.js
RichardBray: Yes you're correct, it does use the user_data to check if someone is logged in or not but your original suggestion for refactoring still valid since I'm disabling the need for a user to be logged in to view the pages.

Sorry I should have made that clearer in the description 🤦

alisalaman-cyb: As discussed, the requirement to be logged in is no longer needed for this section, hence, it's been removed.

PR #365, opened 2022-04-11
Update node-14 branch with latest changes on demo

RichardBray:
# Description

This PR combines the changes that have already been reviewed by CYB devs on the `demo` branch, with the changes on the `feature/node14` branch.

I have no doubt this will need to be updated at some point in the future, but it's easier to work with something that is already in progress than from scratch.

## How to review this PR

https://user-images.githubusercontent.com/1377253/164518471-0f1af813-7ae9-4cb5-b13c-df8706a72346.mp4


api/controllers/ApplicationTypeController.js
```diff
@@ -243,69 +248,84 @@ module.exports = {
...
         let selectedServiceType = req.param('app_type_group');
 
         if (typeof selectedServiceType === 'undefined') {
             selectedServiceType = 'not ok';
         }
 
-        const premiumOrDropoffPageSelected = ['2','3'].includes(selectedServiceType);
+        const premiumOrDropoffPageSelected = ['2', '3'].includes(
+            selectedServiceType
+        );
```
PaulioRandall: `2 == 'premium type'` and `3 == 'drop off type'`?

Sorry, I'm just allergic to magic numbers.

api/controllers/ApplicationTypeController.js
```diff
@@ -243,69 +248,84 @@ module.exports = {
...
+                                'ID already taken, redirecting back to start page'
+                            );
 
                             res.redirect('/start');
 
-                            throw "ID already taken";
-
-                        } else
-                        {
+                            throw 'ID already taken';
```
PaulioRandall: Wouldn't wrapping this in an `Error` be more helpful when debugging in future or does it not make a difference in this instance?

PaulioRandall (commented):
Focused only on your (Rich B) changes. Quite a lot to go through, LGTM but left a couple of non-blocking comments.

api/controllers/ApplicationTypeController.js
RichardBray: That's fair, it does look cleaner without magic numbers 🪄

api/controllers/ApplicationTypeController.js
RichardBray: Good spot, yes this should be wrapped around an Error object 👍

PaulioRandall (approved):
LGTM

PR #369, opened 2022-04-13
feat: add skip questions page with tests

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-458

The purpose of this ticket is to add a skip page (image below) before the suitability questions:

Screenshot 2022-04-13 at 16 01 32

You can view the prototype flow here:
https://fcdo-e-apostille-prototype.herokuapp.com/application/3a-before-you-apply
_I can slack the creds to you_

This was done because people were missing the skip link on one of the eligibility questions.

If you want to run through this code yourself please let me know and I can spin it up on AWS for you. As long as it's after 9am and before 6pm 😉


api/controllers/ApplicationTypeController.js
```diff
@@ -325,7 +325,7 @@ module.exports = {
                                     }
 
                                     if (req.session.appType === 4) {
```
PaulioRandall: I know this is not part of the change but I see magic numbers, what is '_appType === 1_' and '_appType === 4_'?

PaulioRandall (commented):
LGTM. It would be good to a quick walkthrough but not essential.

api/controllers/ApplicationTypeController.js
RichardBray: This is a fair spot, I'm happy to change this.

PR #370, opened 2022-04-14
Remove upload button

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-462

The purpose of this PR is to allow the user two upload files in one step instead of two. 

This removes the need for the user to press the 'Upload' button after they have selected their files only when JS is enabled.

https://user-images.githubusercontent.com/1377253/164038927-6ebc32a2-5bb6-4708-b2cc-50345d565bf2.mp4


assets/js/main.js
```diff
@@ -520,107 +520,13 @@ function setBackLink(){
...
+if (document.querySelector(".js-file-auto-upload")) {
+
+  // Hide the file upload button
+  document.querySelector(".js-trigger-progress-bar").style.display = "none";
+
+  // Automatically upload file on selection
+  document.querySelector(".js-file-auto-upload").onchange = () => {
+    document.querySelector(".js-trigger-progress-bar").click();
+  };
+}
```
RichardBray, line 532: This is the main change

assets/js/main.js
```diff
@@ -520,107 +520,16 @@ function setBackLink(){
...
+
+  if (fileUploadInput) {
+    fileUploadButton.style.display = "none";
+
+    // Automatically click upload button when file is selected
+    fileUploadInput.onchange = () => {
+      fileUploadButton.click();
+    };
+  }
+})();
```
RichardBray, line 535: This is the main change

alisalaman-cyb: @RichardBray does this work on a mobile device?  Worth checking the click function particularly - this post suggests it will but worth a check: https://stackoverflow.com/questions/37273142/would-onclick-event-work-on-touch-on-touch-screen-devices.  

Also, is there a fallback when JS is disabled or is it dependent on callign the `hideUploadFileButton()` function?

PaulioRandall: What's the rest of the app like in terms of no JS?

RichardBray: > @RichardBray does this work on a mobile device? Worth checking the click function particularly - this post suggests it will but worth a check: https://stackoverflow.com/questions/37273142/would-onclick-event-work-on-touch-on-touch-screen-devices.
> 
> Also, is there a fallback when JS is disabled or is it dependent on callign the `hideUploadFileButton()` function?

Hey @alisalaman-cyb thanks for sharing the link. Luckily it all works on mobile 🎉 

Screenshot 2022-04-19 at 16 41 41

When JS is disabled there will be an Upload button that the user has to press to upload the file. The `hideUploadFileButton()` hides that button.

Screenshot 2022-04-19 at 16 44 35

RichardBray: > What's the rest of the app like in terms of no JS?

Hey @PaulioRandall The whole app works fine if JS is disabled actually. The server side code takes priority and the client side JS is just sprinkled on top only when needed.

PaulioRandall: Nice, LGTM.

alisalaman-cyb (commented):
LGTM

PR #377, opened 2022-04-22
Fix/eia 467 casebook 500 errors

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-467

The purpose of this PR is to create a more appropriate error screen for when a user is unable to view an application because of a 500 error from Casebook.

When a user tries to view an e-application from the dashboard, a call is made to retrieve data from Casebook.

Previously if there were any issues getting data from Casebook and a 500 error was received, the user was presented with the page below:
127bd4e0-953c-40a6-933a-f02db1777bf7

The updated error screen in this PR looks like this:
application-a-d-22-0212-1234-abcd


PaulioRandall (commented):
LGTM

PR #379, opened 2022-04-26
Feat/eia 458 add skip page

RichardBray:
# Description
The purpose of this PR is to address to issues Khurram raised.

https://consular.atlassian.net/browse/EIA-458?focusedCommentId=14587

1 - A user is asked to sign in each time they skip the suitability questions.

When a user clicked on the NO option on this page:

Screenshot 2022-04-26 at 15 50 27

They were always asked to sign in even if they have already signed in. That issue has been fixed on this PR.

2 - If you skipped the suitability questions in the above page you would be redirected to the upload page below:

Screenshot 2022-04-26 at 15 53 04

If you then click on the back button from this page it would take you back to the wrong page. This has also been fixed in this PR.

Please let me know if you would like to check these changes on the site itself and I can try and arrange that.


PaulioRandall (approved):
LGTM

PR #381, opened 2022-05-04
EIA-476 Investigate why adding eApp application is not showing or showing twice on the dashboard

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-476
The purpose of this PR is to update the user_id value in the database after the user has logged in.

There is a good explanation of what is going on here:
https://github.com/UKForeignOffice/loi-application-service/blob/e2a0c04c69068e627bb657192712fe7a689d3d8b/docs/eApp-pre-sign-in.md

But if you want the TL:DR version, basically eApp has a feature where you can start the process before signing in but you have to sign in eventually, (on the document upload page).

If a user hasn't signed in they get a user_id of 0 in the database. After they've signed in they get a number like 821 or something in the session, but this isn't added to the db, therefore the application they've started will never show up in their dashboard after they complete it.

This aims to fix that by re-writing the user_id in the DB

I hope that made sense. As usually, feel free to ask questions in slack or as a comment here 👍

### DIFF
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -27,7 +27,7 @@ const POST_UPLOAD_ERROR_MESSAGES = {
 };
 
 const FileUploadController = {
-    setupMulterMIddleware() {
+    setupMulterMiddleware() {
         const multerOptions = {
             storage: uploadFileToStorage(s3BucketName),
             fileFilter: checkTypeAndDuplication,
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -1,11 +1,13 @@
 // @ts-check
 const multer = require('multer');
 const sails = require('sails');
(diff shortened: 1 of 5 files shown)


api/controllers/FileUploadController.js
```diff
@@ -25,26 +27,27 @@ const POST_UPLOAD_ERROR_MESSAGES = {
 };
 
 const FileUploadController = {
-    multerSetup() {
+    setupMulterMIddleware() {
```
PaulioRandall: Very very minor issue but I notice there's a stray capital `I` in `setupMulterMIddleware`. No need to fix but now I've pointed it out it might eat away at your consciousness.

PaulioRandall (approved):
LGTM other than the typo

api/controllers/FileUploadController.js
RichardBray: Ah thanks for pointing this out. I will make this change before merging.

PR #382, opened 2022-05-04
EIA-479 Re-ordering application list removes the application list footer

RichardBray:
# Dashboard
https://consular.atlassian.net/browse/EIA-476

The purpose of this PR is to fix an issue where reordering data on the application dashboard hides the pagination at the bottom.

bdd895eb-a138-4e8b-a81e-d0de057e3e19

The reason this happened was because reordering the application triggered only a partial page to render instead of the whole dashboard, so to fix it the pagination just needed to be added to the partial.


PaulioRandall (approved):
LGTM

PR #383, opened 2022-05-06
Fix/eia 472 incorrect redirect

RichardBray:
# Description
https://consular.atlassian.net/browse/EIA-472

This PR is actually **10x bigger** than it needs to be. 

The main issue it solves is making it so that:
1. when you are completing an eApp before signing in
2.  you get to the stage where you are prompted to sign in
3. you forget or input wrong creds before signing in
4. you aren't redirected to the wrong page.

### The issue is caused by a query param. 

So when you are redirected to sign in mid eApp flow, there is a param `next=continueEApp` which is set to TRUE in the session. 

If you type an incorrect username or password the page refreshes, and that query param disappears from the url, and are taken to the wrong page after you login.

This fix was simply a case of moving some lines of code up and that's it. The bulk of these changes are refactoring so I could write some unit tests. 

*NOTE: The commits themselves might be helpful to go through in this case.*

## Before

https://user-images.githubusercontent.com/1377253/167178059-e6a55f96-b1ae-4b76-bf84-91335737aef5.mp4

## After

https://user-images.githubusercontent.com/1377253/167178102-69c34280-b745-4cca-8cb9-c2fea01f8f75.mp4


tests/specs/controllers/Auth.test.js
```diff
@@ -6,16 +6,16 @@
...
 describe('AuthController:', () => {
-    /* FUNCTION: loadDashboard ---------------------------------------------------------
+    /* FUNCTION: fromSignInPage ---------------------------------------------------------
      *
      */
-    //describe.skip('[Function: loadDashboard]', function() {
+    //describe.skip('[Function: fromSignInPage]', function() {
     //    it('should not error', function (done) {
     //
     //        async.series({
```
RichardBray, line 21: Unfortunately Kainos want this commented out code left in :(

In case you were wondering

PaulioRandall (approved):
LGTM

PR #384, opened 2022-05-09
fix: address accsessibility issues on steps page

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-459?focusedCommentId=14692

The purpose of this PR is to address some accessibility issues raised by Khurram. Here are the issues that have been fixed:

1. ID's with similar names
2. Buttons without text

*There are other accessibility issues but I am unable to fix them since they are part of the layout of the whole application. These might have to be addressed in a separate PR but they are not as important as the ones I have fixed.*


PaulioRandall (approved):
LGTM

PR #386, opened 2022-05-13
Feat/eia 471 improve virus tests

RichardBray:
# Description

This PR predominantly adds tests to make sure that if virus and file type checking was removed then the tests would fail.

It also fixes an issue where file checks where happening on previously uploaded files

Screenshot 2022-05-13 at 15 20 16


tests/specs/controllers/FileUpload.test.js
```diff
@@ -2,15 +2,23 @@ const request = require('supertest');
...
+
+const expect = chai.expect;
 const FileUploadController = require('../../../api/controllers/FileUploadController');
 const HelperService = require('../../../api/services/HelperService');
 const Application = require('../../../api/models/index').Application;
 
 const sandbox = sinon.sandbox.create();
 
+function assertWhenPromisesResolved(assertion) {
+    setTimeout(assertion);
```
PaulioRandall, line 19: Isn't setTimeout supposed to have a time in ms as the 2nd argument?

PaulioRandall (commented):
Just the one minor thing

tests/specs/controllers/FileUpload.test.js
RichardBray: Not necessarily, the second argument is set to 0 by default
https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#parameters

This is sort of a hacky way to trick the runner to run this assertion after the other functions have been executed in the call stack.

PaulioRandall: Ah, a deferred execution. Nice trick.

PaulioRandall (approved):

PR #389, opened 2022-05-17
Fix/EIA 501 - Issue with uploading more than 50 files freezing page

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-501

The purpose of this ticket is to address a bug where the site would freeze when more than 50 files were uploaded. 50 files is the file upload limit but that can be changed in the .env config.

The reason for the freeze was that the page was not being refreshed after the error was logged. Since this is a server-side rendered site, the page needs to refresh for the user to see an update. The error was created for the user to see, but the page was not refreshing. This should be fixed now.

## Also...

I noticed it's possible to upload more than the file limit, then navigate using the address bar to the summary page and complete the application.

Ideally a user would delete the files they didn't want until they got to 50 or below and then proceed, but they could hack it. 

I've added a fix which shows a server error if they progress and exceed the file limit, so they'll have to go back and fix that.

### As usual...
Happy to answer any questions and if you want screenshots and stuff, please let me know 🙏

### DIFF
--- a/api/controllers/CheckUploadedDocumentsController.js
+++ b/api/controllers/CheckUploadedDocumentsController.js
@@ -8,6 +8,10 @@ const addUserDataToDB = require('../helper/addUserDataToDB.js');
 
 const CheckUploadedDocumentsController = {
     async renderPage(req, res) {
+
+        if (HelperService.maxFileLimitExceeded(req))
+            return res.serverError('maxFileLimitExceeded');
+
         const userData = HelperService.getUserData(req, res);
         const { uploadedFileData } = req.session.eApp;
         const totalDocuments = uploadedFileData.length;
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -1,10 +1,8 @@
 // @ts-check
 const multer = require('multer');
 const sails = require('sails');
-const uploadVariables = require('../../config/environment-variables').upload;
+const { s3_bucket: s3BucketName, max_files_per_application: maxFiles } = require('../../config/environment-variables').upload;
 const Application = require('../models/index').Application;
-const { s3_bucket: s3BucketName, max_files_per_application: maxFiles } =
-    uploadVariables;
 const uploadFileToStorage = require('../helper/uploadFileToStorage');
 const deleteFileFromStorage = require('../helper/deleteFileFromStorage');
 const HelperService = require('../services/HelperService');
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -1,7 +1,7 @@
 // @ts-check
 const multer = require('multer');
 const sails = require('sails');
-const { s3_bucket: s3BucketName, max_files_per_application: maxFiles } = require('../../config/environment-variables').upload;
+const { s3_bucket: s3BucketName, max_files_per_application: maxFileLimit } = require('../../config/environment-variables').upload;
 const Application = require('../models/index').Application;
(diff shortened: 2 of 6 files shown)


api/controllers/CheckUploadedDocumentsController.js
```diff
@@ -8,6 +8,10 @@ const addUserDataToDB = require('../helper/addUserDataToDB.js');
 
 const CheckUploadedDocumentsController = {
     async renderPage(req, res) {
+
+        if (HelperService.maxFileLimitExceeded(req))
+            return res.serverError('maxFileLimitExceeded');
```
alisalaman-cyb, line 13: does this error get rendered as a human-readable message for the user in the UI?

api/controllers/FileUploadController.js
```diff
@@ -93,6 +93,17 @@ const FileUploadController = {
         }
     },
 
+    _maxFileLimitCheck(req) {
+        const totalFilesUploaded = req.session.eApp.uploadedFileData.length;
```
alisalaman-cyb: Seems like you should be reusing HelperService.maxFileLimitExceeded here

api/controllers/FileUploadController.js
```diff
@@ -93,6 +93,17 @@ const FileUploadController = {
...
+
+        if (totalFilesUploaded > maxFileLimit) {
+            req.flash('genericErrors', [
```
alisalaman-cyb: is the user prevented from proceeding with the form if genericErrors is populated and if so, where?

alisalaman-cyb (commented):
A few comments for clarification.

api/controllers/CheckUploadedDocumentsController.js
RichardBray: For this specific scenario no, it renders a server error page purely because a user would have to essentially hack the flow to get this error.

Also, Liz is a bit swamped at the moment and wouldn't have any time to write copy for an error page that tells the user to go back and fix the issue they avoided on the upload page 😁

api/controllers/FileUploadController.js
RichardBray: Yes the user is prevented from proceeding, there is [some code in the markup](https://github.com/UKForeignOffice/loi-application-service/blob/develop/views/eApostilles/uploadFiles.ejs#L184) to hide the 'Continue' button on the upload documents page until the user removes the files they don't want uploaded.

api/controllers/FileUploadController.js
RichardBray: This is a far point. The reason this function does a similar thing is to trigger the error message to the user but I can incorporate the ` HelperService.maxFileLimitExceeded` function into this one.

api/controllers/FileUploadController.js
alisalaman-cyb: Thanks.  Correct me if I'm wrong but this coupled with the check in EAppReferenceController this covers the UI (disabling the button) and a bot/hack submitting the form data to the /additional-reference endpoint to attempt to progress the application without having uploaded a valid number of docs, right?

api/controllers/CheckUploadedDocumentsController.js
alisalaman-cyb: No worries.  One to add to the backlog as it should ideally display a sensible error message.

api/controllers/FileUploadController.js
RichardBray: yes that's correct

alisalaman-cyb (approved):
LGTM

PR #390, opened 2022-05-17
Added row describing _crsf on cookie-policy

jonathanhu115:
Adding text to describe the role of the _crsf cookie.


simonmoore100 (approved):
LGTM

PR #392, opened 2022-05-23
Fix/EIA - 504 payment receipt filename

RichardBray:
# Description

https://consular.atlassian.net/browse/EIA-504

The purpose of this ticket is to rename the the downloaded file and add `LegalisationPaymentReceipt-` to the file name.

 
Screenshot 2022-05-23 at 12 26 11

### DIFF
--- a/views/eApostilles/openEApp.ejs
+++ b/views/eApostilles/openEApp.ejs
@@ -148,5 +148,11 @@
 
         <p>Your payment will appear on your bank statement as FCDO Legalisation Service.</p>
 
-        <a href="/download-receipt/<%= applicationId %>" class="govuk-link govuk-!-margin-top-2" role="button" download="<%= applicationId %>.pdf">Download your receipt</a>
+        <a
+            href="/download-receipt/<%= applicationId %>"
+            class="govuk-link govuk-!-margin-top-2" role="button"
+            download="LegalisationPaymentReceipt-<%= applicationId %>.pdf"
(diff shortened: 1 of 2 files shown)


views/eApostilles/openEApp.ejs
```diff
@@ -148,5 +148,11 @@
 
         <p>Your payment will appear on your bank statement as FCDO Legalisation Service.</p>
 
-        <a href="/download-receipt/<%= applicationId %>" class="govuk-link govuk-!-margin-top-2" role="button" download="<%= applicationId %>.pdf">Download your receipt</a>
```
alisalaman-cyb, line 151: Is there validation in the download-receipt controller that processes the download requests to check the user has access to the `applcationId` they're trying to access if they were to hack the URL?

RichardBray: Fair question. Yep there's validation to check the user_id in the session matches the user_id of the application in the database.

Here is where the code for it is https://github.com/UKForeignOffice/loi-application-service/blob/demo/api/controllers/OpenEAppController.js#L235

Since the receipt is straight from casebook and not stored in the application db this is the only way I could think of making the check 🤷

alisalaman-cyb: Thanks for confirming.  LGTM

alisalaman-cyb (approved):
LGTM

PR #398, opened 2022-05-27
Feat/eia 488 character restrictions

RichardBray:
# Description

The purpose of this ticket is to show an error message to the user when an illegal character is used in the optional reference input. This is the only input that exists in the whole of the eApp flow.

## Screens
Screenshot 2022-06-13 at 16 57 55


api/services/HelperService.js
```diff
@@ -850,6 +850,25 @@ var HelperService ={
         }).format(number);
     },
 
+    checkForIllegalCharacters(string) {
+        if (!string) {
+            sails.log.error('checkForIllegalCharacters: string argument is required');
```
alisalaman-cyb: I think this is just logged to the console but does this message get sent back to the user?  A more user friendly message would be recommended here if it is.  Terms like string argument suggest a tech context.

api/services/HelperService.js
```diff
@@ -850,6 +850,25 @@ var HelperService ={
...
+            return {
+                exist: null,
+                characters: [],
+            };
+        }
+
+        const illegalCharacters = new RegExp(/[|&;$%@"<>()+,]/g);
```
alisalaman-cyb: I'd recommend having a list of accepted characters rather than a list of invalid ones.  This way any other character in the char set that you may not have captured, or any new ones that slip in at a later date, can be validated.

views/eApostilles/additionalReference.ejs
```diff
@@ -38,12 +40,12 @@
                     You can also add your own reference here if you want to.
```
alisalaman-cyb, line 40: Would it be worth putting some copy here to explain the characters that are  valid and accepted?

alisalaman-cyb (commented):
A few points added for consideration, most likely will require changes but you may have a valid explanation for some of them.

api/services/HelperService.js
RichardBray: This is a good point. Since this is a helper function, I don't want to include page redirection logic so if there's an error it just returns a null. I don't have any logic on the function that uses this that handles null errors so I'll need to look into that 🤔

views/eApostilles/additionalReference.ejs
RichardBray: That's a good point I will speak to Liz about this.

api/controllers/EAppReferenceController.js
```diff
@@ -19,25 +19,76 @@ const EAppReferenceController = {
...
 
         req.session.eApp.userRef = userRef;
         return res.redirect('/check-uploaded-documents');
     },
+
+    _checkReferenceForErrors({ req, isOverCharLimit, illegalCharacters }) {
+        const errorMessages = {
+            illegalCharacter: {
+                title: 'There is a problem with your reference',
+                text: 'The reference cannot use the following characters:',
```
alisalaman-cyb, line 61: As you've now got a restricted character set you may want to consider including the accepted characters in the error message to give the user an idea of what value they can submit rather than just highlighting the characters that are causing the error.  Not a show stopper but maybe one for Liz to opine on.

alisalaman-cyb (approved):
LGTM.  Just a point on the error message you may want to run past Liz.

PR #400, opened 2022-05-30
refactor: remove unnecessary progress bar code

RichardBray:
# Description

The purpose of this PR is to remove the obsolete code in the upload file progress bar.

The progress bar animation had to be removed because:
- it didn't work well in safari
- it didn't work will with the updated version of Sailsjs

Because of that the only thing that is shown is the stripy bar animation whenever a file is uploaded.

Screenshot 2022-05-30 at 12 57 26

## Also...

There is another change in this PR but I've details that in a commit message so please feel free to refer to that during your review.


PaulioRandall (approved):
LGTM. A good move I feel.

PR #402, opened 2022-05-30
Fix/eia 477 changing url

RichardBray:
# Description
This PR fixes an issue where a user who lands on the service selection page, changes the url to skip the suitability questions, the site prevents them from progressing.

You can go through the prototype to see how the flow works: https://fcdo-e-apostille-prototype.herokuapp.com/application

The reason it breaks is because the a basic application is created in the db at the service selection page and if that is skipped there is no appId to refer the application to. Because the page the user changes the url to in this case will be an eApp specific one, this PR creates an eApp application in the DB when the user clicks on the CTA from the page in question.

https://consular.atlassian.net/browse/EIA-477

### DIFF
--- a/api/controllers/ApplicationTypeController.js
+++ b/api/controllers/ApplicationTypeController.js
@@ -149,7 +149,7 @@ const ApplicationTypeController = {
 
         if (!chosenService) {
             sails.log.error('No service selected');
-            req.flash('serviceSelectError', 'You must select a service type.');
+            req.flash('serviceSelectError', 'You must select a service.');
 
             return res.redirect('/select-service');
         }
--- a/api/controllers/EAppStartPageController.js
+++ b/api/controllers/EAppStartPageController.js
@@ -24,21 +24,24 @@ const EAppStartPageController = {
         const appDetailsInSession = appIdExistsInSession && appTypeExistsInSession;
 
         if (!appDetailsInSession) {
-            await EAppStartPageController._createNewApplicationInDB(req);
+            const newAppData = await EAppStartPageController._createNewApplicationInDB();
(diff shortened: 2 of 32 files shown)


alisalaman-cyb (commented):
The logic looks good to me but one thing to note is that you may want to consider using multiple handlers on the /upload-files path, one that does the checks you're applying with startElectonicApplication() and another being the existing handler you use to process the files.  Not mandatory, just something to consider as I personally find it easier to map the middleware validation that's being applied to a route this way rather than redirecting to different routes.

https://stackoverflow.com/questions/31928417/chaining-multiple-pieces-of-middleware-for-specific-route-in-expressjs

route handlers:
https://expressjs.com/en/guide/routing.html

alisalaman-cyb (approved):
Approving this PR as the logic is fine.  Just a recommendation to consider per my previous comments.

RichardBray:
> Approving this PR as the logic is fine. Just a recommendation to consider per my previous comments.

Hey Ali, this is a far point. I did look into your recommendation and I should have mentioned why I didn't choose with it. Although SailsJS is built on top of ExpressJS it doesn't quite work in the same way when it comes to middleware. All middleware is declared in one file and run before the main application itself. So any arguments that are needed for middleware can't be situational, for example after logging in.

Since the change only applies to people who have gone through the eApp flow before logging in there is no way with the Sails middleware to check if someone has logged in or not because they will always not be logged in since the middleware runs before the application itself. I guess middleware could be written in such a way that it returns a function which accepts specific arguments but I found it easier to put this sort of logic in a helper rather than middleware.

I hope this makes sense.

PR #403, opened 2022-05-30
UAT to Develop

RichardBray:


api/controllers/ApplicationTypeController.js
```diff
@@ -61,179 +50,192 @@ module.exports = {
...
         req.session.selectedDocuments = {
             totalDocCount: 0,
-            documents: []
+            documents: [],
         };
+        req.session.appId = false; // reset the appId so a new session is used
+        // set initial submit status to false, meaning it application has not yet been submitted
+        req.session.appSubmittedStatus = false;
 
-      if (req.query.newSession){
```
simonmoore100, line 71: Not sure why you are removing this section as it was to fix user's who got stuck after logging out?
Unless you are doing something different as I see now when you log out it always takes you back to the /select-service page instead of presenting the login page again

api/controllers/ApplicationTypeController.js
```diff
@@ -61,179 +50,192 @@ module.exports = {
...
+        try {
+            const userDataFromDB = await userModels.User.findOne({
+                where: { email: req.session.email },
+            });
+            const accountDataFromDB = await userModels.AccountDetails.findOne({
+                where: { user_id: userDataFromDB.id },
+            });
+            const standardAppCountQuery =
+                'SELECT count(*) FROM "Application" WHERE "user_id" =:userId and "serviceType" = 1 and "createdAt" > NOW() - INTERVAL \'' +
+                sails.config.standardServiceRestrictions
```
simonmoore100: Mentions to the standardServiceRestrictions was removed as part of the node upgrade.
This looks to be added in again but refactored. Its no longer needed.

Historically, this was introduced a few years ago but didnt really work the way the office wanted it to, so it was abandoned.

api/controllers/FileUploadController.js
```diff
@@ -1,81 +1,183 @@
...
-            fileFilter: checkTypeSizeAndDuplication
-        };
+    async _addSignedInIdToApplication(req, res) {
+        try {
+            const PRE_SIGNED_IN_USER_ID = 0;
+            const { appId } = req.session;
+            if (!appId) throw new Error('No application id found in session');
 
-        return multer(multerOptions).array(FORM_INPUT_NAME);
+            const userId = req.session.user.id || req.session.accunt.user_id;
```
simonmoore100: typo? account?

api/helper/uploadedFileErrorChecks.js
```diff
@@ -293,9 +365,18 @@ function formatFileSizeMb(bytes, decimalPlaces = 1) {
     return `${(bytes / 1_000_000).toFixed(decimalPlaces)}Mb`;
 }
 
+class UserAdressableError extends Error {
```
simonmoore100: Typo? UserAddressableError maybe?

config/security.js
```diff
@@ -0,0 +1,3 @@
+module.exports.security = {
```
simonmoore100, line 1: This file doesn't exist in the develop branch. It's possible this has been re-added when merging the sails upgrade code?

config/cors.js
```diff
@@ -1,78 +0,0 @@
-/**
```
simonmoore100, line 1: This file exists in the develop branch. It's possible this has been removed when merging the sails upgrade code?

config/session.js
```diff
@@ -24,7 +24,7 @@ module.exports.session = {
   pass: session.password,
   db: 0,
   prefix: session.prefix,
-  key: session.key,
```
simonmoore100, line 27: I don't think this needs renamed. It's currently in the develop branch so no need to rename it

config/views.js
```diff
@@ -30,7 +30,7 @@ module.exports.views = {
      *                                                                           *
      ****************************************************************************/
 
-    engine: 'ejs',
+     extension: 'ejs',
```
simonmoore100, line 33: Do you need to change this?

database_change_log/EIA-435_depete_eApp_flag.sql
```diff
@@ -0,0 +1,3 @@
+-- Run against FCO-LOI-Users
```
simonmoore100: Can you rename this file so it follows the naming convention of the others inside the change log directory?
Adding a date to the filename makes it easier to work out at a glance when scripts were created and hence the order they were applied.

views/500.ejs
```diff
@@ -1,3 +1,5 @@
+<% pageTitle="This page cannot be found" %>
```
simonmoore100, line 1: I think this page title is ok for the 404 page but would you need someting different for the 500 page?

simonmoore100, line 1: I think this page title is ok for the 404 page but would you need someting different for the 500 page?

views/dashboard.ejs
```diff
@@ -1,35 +1,52 @@
-<% pageTitle = "My legalisation account - Get a document legalised – GOV.UK" %>
```
simonmoore100, line 1: I think we have a return of the issue we had before whereby if CASEBOOK is down, the dashboard will not load for users.

I need to logon to the Kainos VPN to access the dashboard app statuses when running my local dev, but I can't access the dashboard at all now when I'm not connected.

simonmoore100 (changes requested):
I've left a few comments throughout, nothing that serious and a few of them I'm sure you'll be able to explain instead of change.

api/controllers/ApplicationTypeController.js
RichardBray: Yes when the user logs out it redirects them back to the /select-service page. I think that change was done a while ago so I can't remember exactly why it was made, the change to redirect to the service page. But if you think it's best to keep this code just in case, I'm happy to put it back.

api/controllers/ApplicationTypeController.js
RichardBray: Ah I wasn't aware of this, don't know how it crept in. I'll get rid of it

api/helper/uploadedFileErrorChecks.js
RichardBray: yep, I write loads of typos 🤦 

Thanks for spotting it

config/security.js
RichardBray: fair point, I'll get rid of this

config/cors.js
RichardBray: You're right this does exist, I'll put it back but it doesn't look like it's doing anything 🤷

config/session.js
RichardBray: there must have been a reason for doing this at the time but I honestly can't remember it

config/views.js
RichardBray: This came up as a debug issue for me when I ran the code, I don't know if it shows for you as well 🤔 
Screenshot 2022-06-10 at 17 55 36

config/session.js
RichardBray: I remember the reason now, it came up as a debug message
Screenshot 2022-06-10 at 17 56 48

config/security.js
RichardBray: Actually, this also came up as a debug message for me. Not sure if it shows on develop 🤷 
Screenshot 2022-06-10 at 17 57 48

config/cors.js
RichardBray: I think this has moved to the saols.config.security file. Don't know if there are any specific cors setting that you think should be in there but I'm happy to revert back to the old way if you think it's best.

database_change_log/EIA-435_depete_eApp_flag.sql
RichardBray: ah fair point, don't know how I missed that, I'll fix this

views/500.ejs
RichardBray: Yes, this will be changed in a future PR, thanks for pointing it out though

views/dashboard.ejs
RichardBray: That's odd, for me this works but it takes an incredibly long time to load the page, like 30-40 seconds. If you think it makes sense to put a timeout or something I'll be happy to create a ticket for it.

api/controllers/ApplicationTypeController.js
simonmoore100: Fine to keep this as long as logging out takes you to the /select-service page.

views/dashboard.ejs
simonmoore100: Yeah a timeout would be good, then just display the 'not available' status' if it exceeds it

config/views.js
simonmoore100: Fair point, no warning message on UAT logs so happy to resolve

config/security.js
simonmoore100: Happy to leave as is since there are not warnings on UAT

views/dashboard.ejs
RichardBray: That's fair. I've put a timeout of 3 seconds for now but if you think it should be longer or shorter please let me know.

simonmoore100 (approved):

PR #406, opened 2022-06-09
fix: add various small fixes to the code

RichardBray:
# Description

The purpose of this PR is to fix a few small typos and address an issue with a file limit error.

I've tried my best to make the code as readable as possible but if any of it doesn't make sense please don't hesitate to ask.

### DIFF
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -39,19 +39,22 @@ const FileUploadController = {
         try {
             const noUploadFileDataExistsInSession =
                 !req.session?.eApp?.uploadedFileData;
+
             if (noUploadFileDataExistsInSession) {
                 req.session.eApp = {
                     ...req.session.eApp,
                     uploadedFileData: [],
                 };
             }
+
             const connectedToClamAV = await connectToClamAV(req);
             const userData = HelperService.getUserData(req, res);
             const displayFilenameErrors = req.flash('displayFilenameErrors');
             const infectedFiles = req.flash('infectedFiles');
             let genericErrors = req.flash('genericErrors') ?? [];
+            let fileLimitError = req.flash('fileLimitError') ?? [];
             let backLink = '/completing-your-application';
-            const maxFileLimitErrorExists = genericErrors.includes(POST_UPLOAD_ERROR_MESSAGES.fileCountError);
+            const fileLimitErrorExists = fileLimitError.length > 0;
 
             if (!connectedToClamAV) {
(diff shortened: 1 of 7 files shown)


api/controllers/FileUploadController.js
```diff
@@ -49,8 +49,9 @@ const FileUploadController = {
             const userData = HelperService.getUserData(req, res);
             const displayFilenameErrors = req.flash('displayFilenameErrors');
             const infectedFiles = req.flash('infectedFiles');
-            let genericErrors = req.flash('genericErrors');
+            let genericErrors = req.flash('genericErrors') ?? [];
```
jenbutongit, line 54: 💪 `??`

api/controllers/FileUploadController.js
```diff
@@ -61,7 +62,11 @@ const FileUploadController = {
                 return res.forbidden();
             }
 
-            FileUploadController._maxFileLimitCheck(req);
+
+            if (!maxFileLimitErrorExists) {
+                FileUploadController._maxFileLimitCheck(req);
+                genericErrors = req.flash('genericErrors');
```
jenbutongit: what happens when there are genericErrors, but you don't have max file limit errors? won't this end up replacing them?

api/controllers/FileUploadController.js
```diff
@@ -49,8 +49,9 @@ const FileUploadController = {
             const userData = HelperService.getUserData(req, res);
             const displayFilenameErrors = req.flash('displayFilenameErrors');
             const infectedFiles = req.flash('infectedFiles');
-            let genericErrors = req.flash('genericErrors');
+            let genericErrors = req.flash('genericErrors') ?? [];
             let backLink = '/completing-your-application';
+            const maxFileLimitErrorExists = genericErrors.includes(POST_UPLOAD_ERROR_MESSAGES.fileCountError);
```
jenbutongit: you have a lot of logic finding/dealing with a specific error. should fileCountError have it's own req.flash key?

jenbutongit (changes requested):

api/controllers/FileUploadController.js
RichardBray: ah that's a fair point, I'll address that

api/controllers/FileUploadController.js
RichardBray: yep I can do that, might fix the issue that you raised in the previous comment as well

jenbutongit (approved):

PR #407, opened 2022-06-10
EIA-480 Update error related copy

RichardBray:
# Description

The purpose of this ticket is to update all the error related copy in the eApp according to this spreadsheet
https://docs.google.com/spreadsheets/d/1L8DO1sLr56dvbPUVCcSHtT4HL8VX3usnuGRwpHgGdhk/edit#gid=0

Don't worry if you don't have access I can share a screenshot or something.

### DIFF
--- a/views/eApostilles/additionalReference.ejs
+++ b/views/eApostilles/additionalReference.ejs
@@ -14,7 +14,7 @@
             <div class="govuk-error-summary__body">
                 <ul class="govuk-list govuk-error-summary__list">
                     <li>
-                        <a href="#user-reference">Your reference must be 30 characters or fewer</a>
+                        <a href="#user-reference">Your reference is too long</a>
                     </li>
                 </ul>
             </div>
--- a/views/eApostilles/uploadFiles.ejs
+++ b/views/eApostilles/uploadFiles.ejs
@@ -114,7 +114,7 @@
             <div class="multi-file__uploaded-files">
                 <% if (displayFilenameErrorsExists) { %>
                     <h2 class="heading-medium nomargintop" id="error-list">
-                        <% if (displayFilenameErrors.length == 1) { %>
+                        <% if (displayFilenameErrors.length === 1) { %>
                             1 file has an error
                         <% } else { %>
                             <%= displayFilenameErrors.length %> files have errors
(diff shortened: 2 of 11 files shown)


views/eApostilles/additionalReference.ejs
```diff
@@ -14,7 +14,7 @@
             <div class="govuk-error-summary__body">
                 <ul class="govuk-list govuk-error-summary__list">
                     <li>
-                        <a href="#user-reference">Your reference must be 30 characters or fewer</a>
+                        <a href="#user-reference">Your reference is too long</a>
```
jenbutongit, line 17: Is the user meant to be able to "recover" from this? I'd argue this would be less useful to the user https://design-system.service.gov.uk/components/text-input/#if-the-number-is-too-high

views/eApostilles/uploadFiles.ejs
```diff
@@ -144,6 +144,15 @@
                                 <%= uploadedFileData.length %> files were uploaded
                             <% } %>
                         </h2>
+                        <% if (filesToDelete > 0) { %>
+                            <p id="file-upload-error" class="govuk-error-message">
+                                <% if (filesToDelete === 1) { %>
+                                    Too many files uploaded. You need to delete 1 file
+                                <% } else { %>
+                                    Too many files uploaded. You need to delete <%= filesToDelete %> files
+                                <% } %>
```
jenbutongit:
```suggestion
Too many files uploaded. You need to delete <%= filesToDelete %>  <%= filesToDelete === 1 ?  'file' : 'files' %>
```

jenbutongit (changes requested):

views/eApostilles/additionalReference.ejs
RichardBray: I see what you mean, The line you're highlighting is just for the header bit, there is more helpful text closer to the input


Screenshot 2022-06-13 at 10 35 54

views/eApostilles/uploadFiles.ejs
jenbutongit: p.s. if there' is a lot of this type of code, you would probably be better off implementing a pluralisation or i18n library. 
then in your template you would just do `i18n("errors.tooManyFiles", numberOfFiles)`

views/eApostilles/additionalReference.ejs
jenbutongit: did this come up in UR? I think "summary" is misleading. typically the error in the summary will match the error for the input. https://design-system.service.gov.uk/components/error-summary

RichardBray: Fair point, I will relay that back to LIz. I don't believe this was an issue in UR, this restriction is more to do with security and the copy change 🤷

views/eApostilles/uploadFiles.ejs
RichardBray: That's a good point. I'll create a ticket for that 👍

PR #409, opened 2022-06-14
Feat/eia 485 open eapp copy

RichardBray:
# Description

The purpose PR is to update the copy for viewing the eApp from the dashboard. It focuses on 2 states:
- When all documents have been rejected (https://fcdo-e-apostille-prototype.herokuapp.com/application/A-D-21-0212-1234-ABCD-completed-all-rejected)
- When some documents have been rejected (https://fcdo-e-apostille-prototype.herokuapp.com/application/A-D-21-0205-0343-5BEC-completed)

### DIFF
--- a/assets/styles/informed.scss
+++ b/assets/styles/informed.scss
@@ -2090,3 +2090,8 @@ caption {
   background-color: #f3f2f1;
   padding: 1rem 1rem 0 1rem;
 }
+
+.govuk-notification-banner--error {
+  background-color: #d4351c;
+  border-color: #d4351c;
+}
--- a/assets/styles/informed.scss
+++ b/assets/styles/informed.scss
@@ -2092,6 +2092,6 @@ caption {
 }
 
 .govuk-notification-banner--error {
-  background-color: #d4351c;
-  border-color: #d4351c;
+  background-color: $error-colour;
+  border-color: $error-colour;
 }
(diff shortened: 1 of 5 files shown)


assets/styles/informed.scss
```diff
@@ -2090,3 +2090,8 @@ caption {
   background-color: #f3f2f1;
   padding: 1rem 1rem 0 1rem;
 }
+
+.govuk-notification-banner--error {
+  background-color: #d4351c;
+  border-color: #d4351c;
+}
```
PaulioRandall: Im surprised there isn't a set of colour variables (colour theme). Not suggesting you create them now.

views/eApostilles/openEApp.ejs
```diff
@@ -5,6 +5,24 @@
         <%- partial ('../partials/inner-header.ejs')%>
     </div>
 
+    <% if (applicationStatus === 'Completed' && allDocumentsRejected || applicationStatus === 'Completed' && someDocumentsRejected) { %>
```
PaulioRandall: Could be shortened to?

```html
<% if (applicationStatus === 'Completed' && (allDocumentsRejected || someDocumentsRejected)) { %>
```

PaulioRandall (approved):
One very very minor thing but nothing stopping integration.

assets/styles/informed.scss
RichardBray: Actually that is a good point. I'll check if there are some colour variables before I merge.

views/eApostilles/openEApp.ejs
RichardBray: Ah nice that does make sense, I'll give it a go

PR #411, opened 2022-06-15
Feat/eia 525 feedback consent

RichardBray:
# Description

The purpose of this PR is is to add the feedback consent form the account table to the eApp application when a user creates an application.

How the current service works is that there is a feedback consent section with radio buttons (yes/no) and whatever the user selects will get added to the application in the DB

We don't have this for eApp, so we're just grabbing the feedback consent from the personal details section (table in DB), and adding it to the application itself. 

Screenshot 2022-06-15 at 11 27 39

In the future we think they might put radio buttons there, but this is good for now.

Let me know if you have any questions on this, I don't know if I've added enough information.

### DIFF
--- a/api/controllers/FileUploadController.js
+++ b/api/controllers/FileUploadController.js
@@ -71,7 +71,7 @@ const FileUploadController = {
                 fileLimitError = req.flash('fileLimitError');
             }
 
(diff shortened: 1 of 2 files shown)


PaulioRandall (approved):
LGTM

PR #412, opened 2022-06-15
Fix/eia 515 multiple submissions

RichardBray:
# Description

The purpose of this PR is to fix a bug where a submitted application will show up more than once on the dashboard.

This happens because a function was adding the data to the db when the summary page loaded.

This PR fixes the issue by checking if the data is already in the database before submitting it.


PaulioRandall (approved):
LGTM

PR #415, opened 2022-06-20
fix: add timestap to css file

RichardBray:
# Description

The purpose of this PR is to prevent the browser caching CSS so that new changes don't require the cache to be removed.

### DIFF
--- a/views/layout.ejs
+++ b/views/layout.ejs
@@ -45,7 +45,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <!--STYLES-->
-    <link rel="stylesheet" href="/styles/importer.css">
+    <link rel="stylesheet" href="/styles/importer.1655727270763.css">
     <!--STYLES END-->
     <link rel="stylesheet" href="/fonts/flaticon.css">
 
--- a/tasks/config/sass.js
+++ b/tasks/config/sass.js
@@ -3,7 +3,8 @@
  */
 
 module.exports = function(grunt) {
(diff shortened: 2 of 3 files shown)


views/layout.ejs
```diff
@@ -45,7 +45,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <!--STYLES-->
-    <link rel="stylesheet" href="/styles/importer.css">
+    <link rel="stylesheet" href="/styles/importer.1655727270763.css">
```
PaulioRandall, line 48: I'm curious about how this works. Is the datetime integer `1655727270763` arbitrary?

PaulioRandall (commented):
LGTM except I don't understand the purpose of the arbitrary integer.

views/layout.ejs
RichardBray: Yeah it's a bit weird and I don't entirely know how it works but I have an idea.

1 - All the css/scss in the public folder is moved to a .tmp folder https://github.com/UKForeignOffice/loi-application-service/blob/develop/tasks/pipeline.js#L56
This is because on every build everything here is wiped and rebuilt.

2 - The code looks for some html comments, in this case `<!--STYLES-->` and `'<!--STYLES END-->'` and injects all the css files into that part of the html file.
https://github.com/UKForeignOffice/loi-application-service/blob/develop/tasks/config/sails-linker.js#L84

But it's a super old way of doing things. I don't think I've seen anything like this before.

PR #419, opened 2022-06-23
Fix/eia 531 email link

RichardBray:
# Description

The purpose of this PR is to let the user know how to view an eApp if they are not signed in.

The issue was a user would click on the direct link to an eApp from their email after it has been legalised but, would be presented with an error because they are not logged in. This PR (with the combination of another), hopes to fix that issue.

https://user-images.githubusercontent.com/1377253/175323562-57c7b863-7651-4512-9af4-e7c208ec3a44.mp4

### DIFF
--- a/tests/specs/controllers/OpenEAppController.test.js
+++ b/tests/specs/controllers/OpenEAppController.test.js
@@ -89,7 +89,7 @@ describe('OpenEAppController', () => {
                 config: {
                     hmacKey: '123',
                     customURLs: {
-                        userServiceURL: 'http://localhost/3000/'
+                        userServiceURL: 'localhost/3000/'
                     },
                     casebookCertificate: '123',
                     casebookKey: '123',
@@ -125,7 +125,7 @@ describe('OpenEAppController', () => {
         await OpenEAppController.renderPage(reqStub, resStub);
 
         // then
-        expect(resStub.redirect.firstCall.args[0]).to.equal('http://localhost/3000//sign-in?eappid=test_unique_app_id');
+        expect(resStub.redirect.firstCall.args[0]).to.equal('localhost/3000/sign-in?eappid=test_unique_app_id');
     });
 
     it('prevents viewing the page if application ref is undefined', async () => {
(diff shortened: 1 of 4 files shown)


tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -88,7 +88,9 @@ describe('OpenEAppController', () => {
             _sails: {
                 config: {
                     hmacKey: '123',
-                    customURLs: '123',
+                    customURLs: {
+                        userServiceURL: 'http://localhost/3000/'
```
PaulioRandall: Is this supposed to be `localhost:3000`?

tests/specs/controllers/OpenEAppController.test.js
```diff
@@ -114,15 +117,15 @@ describe('OpenEAppController', () => {
...
+    it('should redirect to sign in page if user is not logged in', async () => {
         // when
         sandbox.stub(HelperService, 'getUserData').callsFake(() => ({
             loggedIn: false,
         }));
         await OpenEAppController.renderPage(reqStub, resStub);
 
         // then
-        expect(resStub.serverError.called).to.be.true;
+        expect(resStub.redirect.firstCall.args[0]).to.equal('http://localhost/3000//sign-in?eappid=test_unique_app_id');
```
PaulioRandall: I notice a `//` at the start of the path. I know it the path gets clean by might be worth fixing.

PaulioRandall (commented):
Just few minor items

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Good spot, I'll sort that out.

tests/specs/controllers/OpenEAppController.test.js
RichardBray: Yeah this could be anything really, it's just a string for unit tests

PaulioRandall (approved):
LGTM

PR #462, opened 2023-02-27
made changes to application type and successful submission views for standard/e-app applications, changes made to live variables which reflect the appropriate estimated processing times for applications

ConorGallagher1:


simonmoore100 (changes requested):
Everything looks fine and works as expected, so good job.

There is just 1 place where the old numOfWorkingDays var is still referenced, index.ejs. This is the landing page when you browse to localhost:3000. The content is out of date here as its not used on production, but you might as well update the var here for completeness.

simonmoore100 (approved):
LGTM

PR #488, opened 2023-05-16
if statement which shows download eapp download link if app status is…

ConorGallagher1:
… complete


simonmoore100 (changes requested):
We just need:

<% if (applicationStatus === 'Completed') { %>

simonmoore100 (approved):
LGTM

PR #491, opened 2023-06-22
Updated sql for exported_app_data function to return only 1 record fr…

ConorGallagher1:
…om each db table.


simonmoore100 (changes requested):
I think the only small change is the 1st line needs to be commented out or it errors out

e.g. -- FUNCTION: public.populate_exportedapplicationdata(integer)

simonmoore100 (approved):
LGTM

PR #508, opened 2024-01-12
FCDO-1743: e-App quirks

simonmoore100:
Render the payment error view when a user tries to submit an app, but their session hasn't been reset. This can happen if they pay successfully and use the back button to submit another app.


ConorGallagher1 (approved):
LGTM

PR #538, opened 2025-03-28
Doc Checker Corrections

ConorGallagher1:
- correcting accept text
- ensuring important notices have the correct design
- ensuring education documents options show the standard cert text
- ensuring emails are in the correct format


simonmoore100 (approved):
LGTM

PR #596, opened 2026-02-26
Update healtcheck route

simonmoore100:
Update healtcheck route so it no longer creates a new redis session on each hit


ConorGallagher1 (approved):
LGTM, approved

